PR #34, opened 2019-09-12
Changed all modpack references to modlist

erri120:
You said yourself that you do not want this to be a modpack installer and rather like the term modlist. I cleaned up all remaining references that contains the keyword modpack and changed it to modlist. This includes not only Strings/Messages but also variables.


Wabbajack/Installer.cs
```diff
@@ -634,7 +634,7 @@ private string HashArchive(string e)
             return HashArchive(e);
         }
 
-        public static string CheckForModList()
+        public static ModList CheckForModList()
```
erri120: In the current HEAD 3e6dd46803c2ad7b040e7b32d6b297cf2d68c969 changed `CheckForModPack` from string to ModList, resolved the conflict for that reason

PR #46, opened 2019-09-19
Improvements to the first startup (fool proofing)

Cyclonit:
**+** Added startup check to determine if LOOT is installed. Without this check, WJ encounters an exception at the very end of the build process.
~ Changed the label for the MO2 profile location selection to "MO2 Profile Location" to clarify what needs to be selected.

Signed-off-by: Christopher Cyclonit Klinge <christ.klinge@web.de>

### DIFF
--- a/Wabbajack/AppState.cs
+++ b/Wabbajack/AppState.cs
@@ -45,12 +45,27 @@ namespace Wabbajack
 
         public AppState(Dispatcher d, string mode)
         {
+            // Ensure WJ is not being run in a download directory.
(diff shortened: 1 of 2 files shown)


erri120:
What if LOOT is installed somewhere else? Maybe open a fileselectionmenu to point to the exe

tstavrianos:
Doesn't wabbajack only care about the loot settings? Those are always in a specific spot in appdata.

erri120:
https://github.com/halgari/wabbajack/blob/master/Wabbajack/Compiler.cs#L142 looks like a user file

tstavrianos:
4 lines above that, it just needs the appdata folder

Issue #49, opened 2019-09-24
OMOD support


erri120:
[This](https://github.com/erri120/wabbajack/tree/omod-extraction-dll) stuff kinda works and uses the dll, had no success with the exe.
Initial tests worked.

erri120:
https://github.com/erri120/wabbajack/commit/498fa8ccf865bce5e0070ccd323a1bfcec72eb3b#diff-53ec1379b88ad1f81b52e3d5eb0ef450R71 needs to changed in final pull request

erri120:
```
0.02 - Wabbajack Build - 04429216f2d9024151bb3269481c30c33e5038d3
0.03 - Looking for attached modlist
0.15 - No Modlist found, running in Compiler mode.
3.54 - Loading VFS Cache
3.58 - Looking for other profiles
3.58 - Using Profiles: Oblivion Test ModList
3.59 - Indexing M:\Projects\Mod.Organizer.2.2.1
3.62 - Updating the cache for 3149 files
4.58 - Extracting ziptestdata.zip
4.6 - Extracting top_level_portion1.zip
4.6 - Extracting ziptestdata.zip
4.62 - Extracting nested_portion1.zip
4.62 - Extracting missing_directory.zip
4.66 - Syncing VFS to Disk: 245 entries
4.68 - Syncing VFS to Disk: 255 entries
4.7 - Syncing VFS to Disk: 262 entries
4.75 - Syncing VFS to Disk: 269 entries
9.29 - Extracting zip_cp437_header.zip
9.3 - Extracting zipdir.zip
9.33 - Syncing VFS to Disk: 1588 entries
15.07 - Extracting DarNified UI 1.3.2.omod
20.54 - Syncing VFS to Disk: 3452 entries
21.31 - Extracting Mod.Organizer.2.2.1.7z
29.47 - Extracting zipdir.zip
29.51 - Syncing VFS to Disk: 6243 entries
29.54 - Extracting zip_cp437_header.zip
30.07 - Syncing VFS to Disk: 6245 entries
30.13 - Extracting missing_directory.zip
30.66 - Syncing VFS to Disk: 6247 entries
30.69 - Extracting nested_portion1.zip
30.72 - Syncing VFS to Disk: 6248 entries
30.75 - Extracting top_level_portion1.zip
30.78 - Syncing VFS to Disk: 6249 entries
30.81 - Extracting ziptestdata.zip
31.39 - Syncing VFS to Disk: 6255 entries
[...]

PR #50, opened 2019-09-24
OMOD Support

erri120:
As shown in #49 works with the two most downloaded omods of all time: [DarNified UI](https://www.nexusmods.com/oblivion/mods/10763) and [DarkUId DarN](https://www.nexusmods.com/oblivion/mods/11280).
Opted to add my repo as a submodule so that we can use the library as a shared project. This means using `git submodules init` and `git submodules update` after cloning the wabbajack repo to initialize and download the submodules.


erri120:
Azure DevOps needs to be told that we have a submodule on our hand or else it will fail.

erri120:
Never used Azure DevOps before but if the commands wont work you may have to adjust the settings somewhere https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/pipeline-options-for-git?view=azure-devops#checkout-submodules

halgari:
fixed and merged, thanks!

PR #52, opened 2019-09-25
Add validation to prevent Wabbajack from crashing when selecting invalid profile directory or starting compilation without specifying profile directory

tr4wzified:
I think it's more user-friendly to notify the user in the log about what they are doing wrong instead of crashing and letting the user find the issue through the log.


tr4wzified:
Checking for another solution where the path will be highlighted in red when the location isn't valid

halgari:
I agree, thanks!

PR #54, opened 2019-09-26
README Updates

erri120:
Contains various Markdown fixes and the addition of recent features.


erri120:
You might also wanna update the Patreons list.

PR #58, opened 2019-09-26
Improved location validation

tr4wzified:
- Location TextBox will now be highlighted in red when an invalid mod profile directory has been selected.
- Location TextBox will get additional tooltip to notify the user about what's wrong with their currently selected location.
- Changed the label that says "MO2 Directory" to "MO2 Profile" to prevent confusion with actual MO2 installation folder.
- Other fields can now also be easily validated by adding them to the switch in the Validate() method (AppState.cs).

### DIFF
--- a/Wabbajack/AppState.cs
+++ b/Wabbajack/AppState.cs
@@ -16,7 +16,7 @@ using Wabbajack.Properties;
 
 namespace Wabbajack
 {
-    internal class AppState : INotifyPropertyChanged
+    internal class AppState : INotifyPropertyChanged, IDataErrorInfo
     {
         private ICommand _begin;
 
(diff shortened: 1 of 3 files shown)


Wabbajack/AppState.cs
```diff
@@ -315,7 +316,46 @@ public string SplashScreenSummary
 
         public void OnPropertyChanged(string name)
         {
-            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(name));
+            if(PropertyChanged != null)
+            {
+                PropertyChanged(this, new PropertyChangedEventArgs(name));
+            }
+        }
+        public string Error
```
Cyclonit: Is this property being used anywhere?

tr4wzified: It's required by the IDataErrorInfo interface.

PR #62, opened 2019-09-29
Minor VFS improvements

Cyclonit:
- Changed deletion of directories in the VFS to not use additional threads, thus improving performance. The only benefit of using additional threads is more information in the log, which the user cannot follow closely anyways.

- Fixed hash-cache-files (.sha) being hashed on new runs.

- Changed the naming of temporary directories of the VFS to not use full GUIDs, as this generates unnecessarily long directory names. Paths with more than 255 characters are not supported on Windows machines. It now uses the thread-id, which is sufficient to avoid threading issues.

### DIFF
--- a/VirtualFileSystem/VirtualFileSystem.cs
+++ b/VirtualFileSystem/VirtualFileSystem.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Threading;
 using Compression.BSA;
 using ICSharpCode.SharpZipLib.Zip;
 using Newtonsoft.Json;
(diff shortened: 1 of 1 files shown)


halgari:
The reason I call the shell command is that we had way too many issues with read only files. Using a shell `del` command we can tell the OS to delete in all situations.

halgari:
Using a ThreadID vs a guid means that if a thread ever calls those functions twice we'll clobber the results from the previous run. 

Windows routines have a 255 character limit which is why we use AlphaFs (look at the imports at the top of the file)

Cyclonit:
At least on my machine, the external threads did not always exit cleanly. Instead the loop reading the output of the deleting thread got stuck waiting for more lines. 

I frequently got "Could not find part of the path" exceptions while using longer path names than 255 characters.

I'll take another look though on both of these issues and add them to github proper.

tstavrianos:
Why not use a static counter variable for the directory names (using it with Interlocked.Increment to make it thread-safe) ?

As for deleting read-only files, what about: https://stackoverflow.com/questions/611921/how-do-i-delete-a-directory-with-read-only-files-in-c

PR #64, opened 2019-10-01
Do VFS directory deletion in existing threads

Cyclonit:
Changed the VFS to delete directories in the main thread instead of spawning a new process running the windows command "del". del was used to deal with read-only files. This can however be done using AlphaFS directly, thus eliminating the need for additional threads.

DirectoryInfo.GetFileSystemInfos() does a breath search. By reversing the result, we are guaranteed to delete from the deepest level of the hierarchy up to the root. Thus directories are always empty once we reach them.

### DIFF
--- a/VirtualFileSystem/VirtualFileSystem.cs
+++ b/VirtualFileSystem/VirtualFileSystem.cs
@@ -1,18 +1,17 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


erri120:
Is this still relevant ?

Cyclonit:
Relevant yes, but I currently don't have any time to work on this. Feel free to close it if it bothers you.

halgari:
We'll close it for now, and perhaps reopen it in the future when we have time to confirm that it all works correctly. In the past I had lots of issues getting C#'s Delete routines to properly delete read-only files. That may have changed since then though.

PR #70, opened 2019-10-05
Fix validation bugs, corrected error in README.md & MessageBox

tr4wzified:
Also fixed the GitHub link to redirect to the new repo.


tr4wzified:
And I added a way to set the label for the location, to prevent users from getting confused when installing and seeing "MO2 Profile:".

PR #73, opened 2019-10-07
UI Stuff

erri120:
Frontend:
- Checkbox to disable NSFW mods in the Slideshow
- Checkbox to disable the entire Slideshow
- Next button to view the next slide
- Scaleable UI

Backend:
- New Modlist Propertysystem:
  - Change Modlist Name, Author, Description and Website
  - Include custom banner
- new Slide precache, cache and queuing system
- UI is now in a UI folder


Wabbajack/AppState.cs
```diff
@@ -29,15 +33,27 @@ internal class AppState : ViewModel, IDataErrorInfo
 
         private readonly Dispatcher dispatcher;
 
-        public AppState(Dispatcher d, string mode)
+        public AppState(Dispatcher d, TaskMode mode)
         {
             var image = new BitmapImage();
             image.BeginInit();
-            image.StreamSource = Assembly.GetExecutingAssembly().GetManifestResourceStream("Wabbajack.banner.png");
+            image.StreamSource = Assembly.GetExecutingAssembly().GetManifestResourceStream("Wabbajack.UI.banner.png");
```
Noggog: Should make use of UIUtils.BitmapImageFromResource

Wabbajack/AppState.cs
```diff
@@ -165,8 +183,57 @@ public ICommand VisitNexusSiteCommand
...
         }
 
-        public string _nexusSiteURL = null;
+        public ICommand OpenModListPropertiesCommand
+        {
+            get
+            {
+                return new LambdaCommand(() => true, () => OpenModListProperties());
+            }
+        }
```
Noggog: Can scrunch this a bit to `public ICommand OpenModListPropertiesCommand => new LambdaCommand(() => true, OpenModListProperties);`

Additionally, maybe if you add a ctor to LambdaCommand that doesn't have a canExecute parameter (eg, internally sets it to always be executable), this could just be
`public ICommand OpenModListPropertiesCommand => new LambdaCommand(OpenModListProperties);`

Wabbajack/AppState.cs
```diff
@@ -175,6 +242,30 @@ private void VisitNexusSite()
             }
         }
 
+        private ModlistPropertiesWindow modlistPropertiesWindow;
+        internal string newImagePath;
```
Noggog: I've noticed `internal` being used around, which is N/A in this project's context
Essentially `internal` means "`public` for this library, `private` for users of this library"

But since we're writing a GUI, there's never going to be any users of this project as a library, so any `internal` are basically just `public`.  Not a gamebreaking issue, but just thought I'd mention it

Wabbajack/AppState.cs
```diff
@@ -262,69 +381,198 @@ private void UpdateLoop()
...
+                            }
+                        });
+                    }
+                    _originalImage = false;
+                    SplashScreenModName = element.ModName;
+                    SplashScreenAuthorName = element.AuthorName;
+                    SplashScreenSummary = element.ModSummary;
+                    _nexusSiteURL = element.ModURL;                  
+                }
+                if (fromLoop)
```
Noggog: Hmm, do we want to check this here?   By only updating the timestamp when it's an update triggered by the loop, you may get a situation where the user manually hits "next", which triggers a flip, but then the loop timer fires right after, leading to a second spammy flip.

I'd think we'd want to reset the timestamp regardless of where the update came from.  Any update, whether auto or manual click should make the system wait the X seconds before doing another auto-flip.

Wabbajack/AppState.cs
```diff
@@ -17,8 +17,12 @@
 
 namespace Wabbajack
 {
+    public enum TaskMode { INSTALLING, BUILDING }
```
Noggog: Super big fan of this enum over the magic strings.  👍

Noggog:
Might be outside the scope of this PR, specifically, but I think this slideshow functionality warrents the creation of a sub class that encapsulates its logic away from the main view model that is AppState.cs.

If we don't start segmenting logic, as more and more functionality is added to Wabba..  AppState will get to be a massive monolith of every last feature we've ever added.

Adding something like SlideshowVM.cs where all the slideshow timers/variables/functions go might help compartmentalize features.  

AppState can then declare `public SlideShowVM Slideshow { get; } = new SlideShowVM();`
Then the XAML in MainWindow can just do stuff like
```
<!-- The datacontext is the AppState.cs, so navigating to Slideshow.SplashScreenImage will access the subVM, and attach to the image -->
<Image Grid.Row="1" Grid.Column="0" Source="{Binding Slideshow.SplashScreenImage}" Stretch="Uniform" Margin="5">
```
to latch onto the sub-VM.

If you wanted to push this even further, could create a new UserControl for the slideshow XAML, too.  Moving lines 36-51 from MainWindow.xaml to its own SlideshowControl.xaml.
Then MainWindow would just have 
```
<!-- The datacontext is the AppState.cs, so setting this control's datacontext to the Slideshow VM object, will mean all the bindings inside SlideshowControl are relative to that sub-VM and don't need the "Slideshow." prefix from the snippet above -->
<local:SlideshowControl DataContext={Binding Slideshow} />
```

PR #76, opened 2019-10-09
ViewModel class /w RaiseAndSetIfChanged

Noggog:
Made a common base class that implements INotifyPropertyChanged.  Also implements a RaiseAndSetIfChanged call that streamlines property definitions.  Some other various cleaning


Noggog:
Snippet-Installation

Little helper vid for those uninitiated with snippets

halgari:
Thanks, looks great!

PR #86, opened 2019-10-12
Adding Rx libraries.  Some initial migration to Rx concepts

Noggog:
This is a commit to add Rx concepts to Wabbajack's GUI.

It adds a few libraries:
- System.Reactive (Rx for C#)
- ReactiveUI (Rx for WPF)
- DynamicData (Rx concepts for collections)

These additions are made in the first commit, which is the only commit that "adds Rx libraries".
The remaining commits are my initial refactoring of the existing systems to Rx patterns.  I felt like just adding the Nugets as a PR wasn't that interesting, and it would be good to demonstrate why these packages were being added in the first place and what they could provide.

I tried to segment each change into a commit that is digestible one-by-one, as the overall change delta might be a bit large.  **If these refactors are too much, I can dial everything back and make a new PR /w just the Nuget additions.**

Here's a summary of the refactors I did:
**- ViewModel now inherits from ReactiveObject**
Its own property changed systems removed.  A convenience CompositeDisposable was added for any Rx subscriptions to be automatically cleaned up when the viewmodel is disposed

**- Swapped all existing commands to ReactiveCommands, deleted LambdaCommand**

**- Removed Dispatcher members in favor of static reference to the dispatcher.**
  This is sort of moot, as most explicit invoke on dispatcher calls should disappear as Rx gets utilized more, anyway.

**- Refactored some commands IsEnabled wiring** 
Changed readme button's IsEnabled functionality to work off of command's CanExecute systems. 
Modified the "Begin" button's executable state to be driven by UIReady property.

**- Slideshow's bootstrapping code that was in ApplyModlistProperties was refactored out.**
  An Rx chain was set up to update the splash image to the wabbajack logo by default, or if the slideshow is disabled or not ready.  When a modlist is set, and its variables are viable, it will automatically swap to the modlist image

**- ApplyModlistProperties() was removed entirely**.  Instead the logic subscribes to the ModList property, and updates the appropriate properties when it changes.

**- The CPU status systems were refactored and simplified.**  The dual lists, dirty property, and thread loop that checked the dirty property and migrated items from the internal list to the GUI bound list has been removed.

Instead, the WorkQueue notifies a subject (Rx's concept of an event invocation) when it has any progress updates.   The updates are subscribed to, sorted by ID, throttled, and then bound to the GUI observable collection.  This is powered by the DynamicData library, which streamlines collection concepts in Rx.

I did bump up the refresh rate to 250 milliseconds (up from once a second).  Can change that at will by modifying `.Batch(TimeSpan.FromMilliseconds(250))`

**- Removed ShowReportButton property in favor of a IsNotNullVisibilityConverter.**  Not related to Rx, really, haha.

**- Refactored slideshow update systems.**
Removed completely the UpdateLoop and associated thread.  
Instead, slideshow images are updated by an Rx chain that triggers off of a timer, and/or the user requesting the next slide.  This is then throttled to prevent spam, and only fires if slideshow is enabled and the program is currently installing.

-----------

There's some more changes I was eying to make, but I thought this was a good foundational baseline to bootstrap the systems to be more Rx.

I tested along the way making these changes by starting up Wabbajack and removing shas and rehashing "Modlist Lite" that I found on the discord.   I didn't wait for it to finish each time.  I just started it up, and confirmed all the buttons still clicked, CPU progress still showed, slideshow flipped, etc.


Noggog:
Going to do another correctness pass on this tonight, as well as add some more comments, probably

halgari:
Thanks for this, I'll be test driving this in a few hours, and merge it once I understand the changes.

PR #88, opened 2019-10-13
Fix for Laptop Screens

giraldiego:
This was the easiest and quicker solution I was able to find using my limited knowledge in wpf and xaml.


Noggog:
I think you mentioned this in the discord, but I don't think this is the proper solution.  Your tweak just means the Viewbox will stretch its contents to be warped, rather than keep the proportions in line.

erri120:
Very bad changes, sry

giraldiego:
I don't know how to use this yet

PR #92, opened 2019-10-13
Xaml IDE featureset

Noggog:
Accidentally closed my previous PR for this.

While developing, I was noticing that a lot of GUI features that VS usually provides me weren't working.  Don't get me wrong... XAML IDE features are VS's weak point by far, but it does have a few things that I haven't been seeing:
**1)  XAML <-> ViewModel intellisense**
If you add `d:DataContext="{d:DesignInstance local:AppState}"` to MainWindow.xaml, for example, it should intellisense all the properties found on AppState.cs when writing Bindings.

XAML VM Intellisense

I fixed this by adding `<XamlDebuggingInformation>True</XamlDebuggingInformation>` to the solution properties, as I guess any build besides the default "Debug" doesn't have this on?

**2) Hot Reload updates**
When a GUI is open in debug mode, you usually are able to make changes in the XAML, save, and see the results immediately in your open running GUI.

XAML Hot Reload

This was a bit more tricky.  I think the reason it's broken is the XAML designer in VS runs in 32bit?  And since Wabba only had 64 bit options, the feature was disabled?

This was the article I found that lead me to the solution of adding 32bit build options.
https://stackoverflow.com/questions/17484109/designinstance-not-working-in-vs2012

When building in 32 bit mode, now, I'm able to get XAML VM intellisense again.  There might be a better solution, but this is what I've come up with.

If we want 64 bit Wabbajack apps, we might just use the 64 bit builds when distributing?  But still have a 32 bit for XAML dev.  Not sure.


Noggog:
Made heavy use of this in #138 
I added it while developing, but removed it before making the PR.  The hot reload systems really are night and day being able to develop and see the changes immediately.

PR #93, opened 2019-10-14
Fix for low resolution screens

giraldiego:
A small fix


Wabbajack/UI/MainWindow.xaml
```diff
@@ -8,8 +8,8 @@
     Title="Wabbajack"
     Width="1280"
     Height="960"
-    MinWidth="1280"
-    MinHeight="960"
+    MinWidth="1024"
```
erri120: Still don' think this is a good idea but will check later

giraldiego: Thank you erri, this way it won't affect the current behavior of the app, but it will work for people working on laptops that don't have full HD displays.

halgari:
I think we need to fix this bug for real, probably by making the UI fully resizable

Noggog:
Yeah.  I do think @giraldiego 's fix is quick and easy.  Maybe just pick some values that have the same proportions as the starting size.  (1024x768)?

Doing that doesn't have any dire downsides that I can think of offhand.  The behavior is the same as someone up-sizing, just in reverse.

If you really don't want it to be a problem, even for someone with an even smaller screen... just go all out and make the minimum very small.  It's the user's prerogative to make it oddly tiny if they want.  All the min needs to be there for is to make it so they can still at least click on it to make it bigger again.  I usually just pick something like 200/300 so it can't just collapse to a pixel

PR #97, opened 2019-10-15
MainWindow spawns on ModeSelectionWindow

Noggog:
Just a quick fix to make the MainWindow spawn wherever the Mode Selection window is at the time of button clicking.   Feels smoother


halgari:
It looks like this is missing the null check on the return value from the open file dialog

Noggog:
It's just hiding in the OpenMainWindow call, L44

PR #111, opened 2019-10-23
Reenabled OMOD extraction

erri120:


halgari:
Thanks, sorry about that.

PR #112, opened 2019-10-23
Include WJ build version in the Modlist

erri120:
* Changed assembly information
* Added a string representation of the `FileVersion` to the modlist properties
* User will be notified via `MessageBox` about version version discrepancy

### DIFF
--- a/Wabbajack.Lib/Compiler.cs
+++ b/Wabbajack.Lib/Compiler.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
+using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
 using VFS;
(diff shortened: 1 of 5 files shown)


Noggog:
I feel like this will fire too aggressively.  If we push a new version of Wabba /w 100% just GUI polish changes, increment the version (because it's a new version), now people will get this popup and start getting paranoid their modlists won't install properly.  Then everyone's requesting the authors rebuild with the new wabbajack, etc.

I think we need more control over when we mark older modlists as non-viable and warn about it.

Either:
1)  We only warn if the major version is older/newer.  This way we can reserve increment that number for when a modlist breaking change occurs.  Problem might occur if we have too many minor pushes w/o introducing a breaking change (255)
2)  We have a separate version stamp for the modlist structure.  We increment this number when we've made a change that affects the compatibility of modlists.   Similar to now where we have `.modlist_v2` file extension... except now it'd just be a field in the json rather than a different extension.
Upside of this is we very explicitly can control when we deem a version breaks compatibility.

halgari:
The code is a bit too aggressive (as mentioned) but that's also something we can back up on a bit. So I'll merge the PR now, and we can switch it to fail less often in a future PR.

PR #133, opened 2019-10-31
WJ will close if MO2 is running

erri120:
Wabbajack will load the entire list of running processes and check if `ModOrganizer.exe` is in there. WJ will also check if `ModOrganizer.exe` is inside the MO2 Folder of the selected modlist.
Only works in compile mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Slideshow timer resets when next is pressed
 * Changed modlist extension to `.wabbajack`
 * You can now open modlists directly (after initial launch)
+* Wabbajack will exit if MO2 is running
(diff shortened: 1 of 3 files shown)


halgari:
Full on app crash seems a tad harsh, but I'll need to update VFS's integration with FS changes before we can take a better approach.

Noggog:
Reposted from the discord:
Yeah, looks like a good check to have.  I was going to suggest migrating a lot of these "there's a problem, close the program then fix it" checks to Rx powered warnings that display on the GUI and block starting any processing until the problem is resolved. (orange errors /w tooltips or whatever)

But I think we can do that a little later

Issue #134, opened 2019-11-02
Vortex Support


erri120:

erri120:
Just something I'm thinking about doing.
Vortex vs MO2:
Vortex is not portable, it comes with an installer and normally get's installed at 
`C:\Program Files\Black Tree Gaming Ltd\Vortex`
User files are stored at `%appdata%\Roaming\Vortex`
Vortex is an electron application, state is managed by redux.
When adding a new game:
Vortex creates a new folder at `%appdata%\Roaming\Vortex\{gamename}`, two new folders will be created in that one called `mods/` and `profiles/`. Deployment files are stored as `.msgpack` (https://msgpack.org/) files.

erri120:
After some testing on my Hyper-V virtual machine:
Vortex needs to be in `Per-User` mode so that the mod staging folder is at `%appdata%\{game}\mods`.
Default deployment method is hardlink deployment.

When clicking downloading with Vortex:
Download goes to `downloads\{game}`. Adding manual downloads to that folder will make them appear in the `Mods` and `Downloads` section in Vortex. In the mods section they will have the status `Never Installed`.

Installing mods in Vortex:
Mods labeld as `Never Installed` will be extracted to the `{staginfolder}\{archivename}\`.
Noteable there will be a `project.xml` file created:
```xml
<?xml version="1.0" encoding="utf-8"?>
<project>
    <PreviewIconFile>preview_icon.png</PreviewIconFile>
    <ItemDescriptionShort/>
    <ModDataPath>c:\program files (x86)\steam\steamapps\common\DarkestDungeon\mods\OksTrinketsvv</ModDataPath>
    <Title>OksTrinketsvv</Title>
    <Language>english</Language>
    <UpdateDetails/>
    <Visibility>public</Visibility>
    <UploadMode>direct_upload</UploadMode>
    <VersionMajor>1</VersionMajor>
    <VersionMinor>0</VersionMinor>
    <TargetBuild>0</TargetBuild>
    <Tags>
    </Tags>
    <ItemDescription>File generated by Vortex</ItemDescription>
</project>
```
When deploying the mod, the `vortex.deployment.msgpack` file gets created/updated and filled with deployment info.
Still need to check if this is game specific.

Installing mods inside the staging folder manually:
This is possible but requires the exact folder structure as told above including the `project.xml` file. Vortex has to be closed during this time and after opening Vortex again, it will complain about manual tempering with the staging folder. Mods can than be deployed.

[...]

PR #142, opened 2019-11-03
Vortex support

erri120:
DO NOT MERGE YET
See #134


erri120:
> 
> 
> I don't like the changes to this and to the other game registry bits. This change goes from game names being a well defined, strongly typed, enum, where WJ is the source-of-truth, to an arbitrary string that must be whatever the Nexus uses. This also merges Skyrim VR into Skyrim SE, simply because the Nexus decided not to split the two apart.
> 
> If we really want to support a lot more games, we need to hand validate each, and add their metadata.

If you want to add all 69 official supported Vortex games + undefined amount of unofficial to the list than by all means go ahead. I removed the `Game` enum because it kinda made no sense to have.
The `GameRegistry.Games` Dictionary will still be populated with the games we support (tested or not). 
The idea is that we have a list of games we know that WJ can somewhat handle. If the user goes ahead and wants do create a Vortex modlist, WJ will check the reg paths of the games in the dictionary to see what Vortex+WJ supported games are installed. The user will still be able to manually choose games we don't have in that list.
Once the game is selected, the staging path + downloads folder will be prompted or WJ will find them if they are at the default locations (`%APPDATA%\Vortex`).

halgari:
I think we absolutely should validate each game by hand, and I don't like the idea that we're tying ourselves to the nexus in this way. And like I said, we're losing SkyrimVR as a discrete entity, as it's now stored simply as "skyrimspecialedition". Speaking of which, how does this compile with a duplicate key? `https://github.com/erri120/wabbajack/blob/vortex-support/Wabbajack.Common/GameMetaData.cs#L115`

Vortex doesn't support so many popular games that it would take a long time to validate each as interest arises. Most games here `https://www.nexusmods.com/` (sort by number of mods) hardly have over a thousand mods, and many (like Mass Effect, and Battletech) don't use Vortex as part of a normal modding experience.

halgari:
There were some interesting points raised by some of the WJ support staff and they are things we should consider before going live with this feature: 

1) Will users understand that this won't allow someone to install a list that was created in MO2 and drop it into Vortex? We all know that the installation manager is hardcoded by the list's author's choice, but how do we communicate that? 
2) MO2 modlists won't require the modlist to be installed first, but Vortex will, this complicates communication as it requires users to do a step before running WJ if they want to use something that uses Vortex. 
3) Support staff for WJ don't use Vortex, and few of them play the non Bethesda games Vortex would be using. How do we support this feature?

erri120:
> 
> 
> There were some interesting points raised by some of the WJ support staff and they are things we should consider before going live with this feature:
> 
>     1. Will users understand that this won't allow someone to install a list that was created in MO2 and drop it into Vortex? We all know that the installation manager is hardcoded by the list's author's choice, but how do we communicate that?
> 
>     2. MO2 modlists won't require the modlist to be installed first, but Vortex will, this complicates communication as it requires users to do a step before running WJ if they want to use something that uses Vortex.
> 
>     3. Support staff for WJ don't use Vortex, and few of them play the non Bethesda games Vortex would be using. How do we support this feature?

1: When they open a `.wabbajack` ModList they will get a message about the Mod Manager used and different settings for installation.

2: Can you elaborate more, I don't know what 'extra steps' you are talking about.

3: The current `VortexCompiler` I create is only for non bethesda games. Since Bethesda games are the main games on NexusMods the userbase on NM that use Vortex for other games is very small. I checked most of the non BGs that Vortex supports and all I looked at are simple extensions that just tell Vortex where to find the game and where to install the mods. No external tools like FNIS, SkyProc, DynDOLOD or all the other tools you'd use when modding BGs.
I only use Vortex for non BGs like Darkest Dungeon, KOTOR, Starbound,.. where you do not need all those extra fancy pants tools but where you just slap different mods together and maybe rearrange the load order. I mean just look at the Compilation Stack for this Compiler:
```cSharp
new IncludePropertyFiles(this),
new IncludeVortexDeployment(this),
new IncludeRegex(this, "^*\\.meta"),
new IgnoreStartsWith(this, " __vortex_staging_folder"),
new IgnoreEndsWith(this, "__vortex_staging_folder"),
new IgnoreEndsWith(this, "project.xml"),
new IgnoreGameFiles(this),
new DirectMatch(this),
new IgnoreGameFiles(this),
new IgnoreWabbajackInstallCruft(this),
new DropAll(this)
```
this is nothing compared to how the normal Compiler looks like.
This VortexCompiler will **never** support BGs and is designed to create simple ModLists and to show NM that Vortex is not their exclusive playground. I will try to give full support for this feature. We'd need new channels for Vortex only stuff.

halgari:
We have two tests that are failing 
test errors

I suggest running the unit tests locally to figure out what went wrong, it looks like maybe a path isn't set properly after the abstraction of ACompiler

PR #143, opened 2019-11-03
update all deps

halgari:


erri120:
Do not update my omod framework

PR #151, opened 2019-11-06
Settings File

Noggog:
Adds a savefile, which saves some settings across different runs. 

wvJuphkK8H

  - Savefile "settings.json" is created and stored next to the exe
  - Saves settings per modlist and per MO2 profile, so installation locations will be remembered even if user installs many different modlists, for example.
  - Right now the syncing systems are very basic.  They have separate lines for putting the settings into the VM, and for getting them from the VM.  This opens up bugs to where one side is forgotten or not maintained symmetrically.   I'd like to eventually upgrade them to a Bind-style setup maybe similar to https://reactiveui.net/docs/handbook/data-binding/windows-presentation-foundation.  Can't use those directly, as they're too tied to UserControls.  Maybe they've got something else similar that I just need to find.

Other changes:
  - Migrated extension functions to Common library
  - Bugfix in FilePicker
  - Extracted XAML needed to support Mahapps watermark systems for textboxes.  This will also help prevent people from accidentally submitting "Description (700 characters max)" as their description.

tiFQqi1hJp


halgari:
Looks great, thanks!

PR #155, opened 2019-11-07
UI Stuff #3

erri120:
- only one Window
- new start screen
- new step/page like navigation
- new gallery
- new config system
- new baseline for more implementation


Wabbajack.Common/Utils.cs
```diff
@@ -27,10 +27,17 @@ namespace Wabbajack.Common
 {
     public static class Utils
     {
-        public static bool IsMO2Running(string mo2Path)
```
Noggog: I like this refactor.  More generic so we can use it for lots of purposes.
Someone can readd an explicit IsMO2Running() /w the parameters hardcoded if they really want to, or we can have the "ModOrganizer" string a static const somewhere for easy reference when using this function

Wabbajack/View Models/CompilerVM.cs
```diff
@@ -2,19 +2,20 @@
...
+using DynamicData;
+using DynamicData.Binding;
 using Wabbajack.Common;
 using Wabbajack.Lib;
 
 namespace Wabbajack
 {
     public class CompilerVM : ViewModel
     {
-        public MainWindowVM MWVM { get; }
```
Noggog: This refactor suggests there was a misunderstanding with what this property was for, and how the Queue/Log systems were organized.

WorkQueue is a static singleton.  Logs that get output are also a static singleton, of sorts.  Any view should be able to get at their results and display them if they want, but don't necessarily need their own copy.  MainWindowVM is/was a good spot to put singleton concepts, as we know there's only going to be one, and it's the parent of all.

This MWVM property was so that CompilerVM/InstallerVM have a way of getting to the MainWindowVM to get at those singleton collections.  It's public (and short) so that View's can use them in bindings.  Assuming a datacontext of CompilerVM, `{Binding MWVM.Logs}` would start the bind at CompilerVM, walk up to the MainWindowVM, and bind to the singleton log list for display.  Any subVM could walk up and reuse the same backing data in their displays.

Wabbajack/View Models/CompilerConfigVM.cs
```diff
@@ -0,0 +1,89 @@
...
+                    MO2Effect = null;
+                    VortexEffect = StdBlur;
+                }
+                else if (VortexEffect == StdBlur && !b)
+                {
+                    VortexEffect = null;
+                    MO2Effect = StdBlur;
+                }
+            }
+        }
```
Noggog: This sort of logic should preferably be powered by the View.  VMs should know as little as possible about the actual end display as they can.  For all they know, 10 different views might be bound to the same VM and be displaying different things.  

As an example, in a game, perhaps, you might have an NpcVM, with health, name, inventory, whatever.  An InventoryView might bind to an NPC, and display its inventory by binding to it, while a separate HealthBarView might also bind to the same object, and display the health.   NpcVM should be unaware that InventoryView/HealthBarView exist, or even care.  100 things might be bound to it and be using some of its data, for all it knows.

So back to our current code here:  The active effect should be managed by some XAML style triggers.  They should bind to some enum "Vortex/MO2", and if vortex, swap to one effect vs if MO2 swap to another. etc.  The fact that an effect even exists or is something that needs to be swapped is very specific to the view itself, and should stay in that realm

Noggog:
### Front Page
I like a good amount of the front page.  Color matching the logo is nice.  Icons at top right look nice.

I'm not a fan of the use of the Tile control itself.  
1)  The very basic black square doesn't really look too nice.  Additionally, there's no visual feedback at all when hovering over to help the user realize it's a clickable item.  I think going back to `<Button>` and just maybe defining a new "large button style" with some tweaks to make it look better for larger buttons would be better.  Not sure what the tiles are for if they have such low interaction included... they're basically just buttons without any hover or click feedback?

2)  I don't like the scrollviewer use.  Gives it a jiggle for no reason.

3iZuc5EcaD

3)  Don't like the items control wraparound tiling setup here.  Should be a grid with three * width'd columns, so the buttons never go onto two lines

notepad++_yswWC49sO3

Overall I think it's a good base and introduction to the app, maybe with a few more styling polish additions.

### Browse Mod Page
I like the tile items control (and by tiles i just mean square objects, I do think these can be buttons to get back some hover styling, too).  The fact that the tiles line up and fit as you resize is a good setup.

I like the back button functionality.  Was something I think the app desperately needed was some two-way navigation.

1)  I think the Back button can be put on the same line as the github/patreon/discord icons.  We're losing a lot of space just for those icons, and putting the back button up there not only reuses that space well, but also makes it always on screen to be clicked.

2)  I think the typical buttons need more margin on them in my opinion.  (I guess in the whole app now?)  Icons rubbing up on the edges like that doesn't look right.  I think "Padding" can just be added to the style.

3)  I think the blur effect wasn't applied properly to the titles.  As they are now it looks like the ticker text you see when you enable subtitles while watching a show.

4)  I'm not sure how I feel about having buttons with both text and icons.  IMO a good icon should communicate the goal of a button, and tooltip will be there to back it up if there's any confusion.  For ex, "Settings" buttons that use the typical gear icon don't write out "settings" alongside the gear.  Everyone knows what it is.  Could be convinced otherwise, and I do think that some more obscure buttons might be worth having text on.

5)  While I appreciate the bold design choice of having the WJ logo large and in charge up front, I think it's getting to be a gimmick at this stage in the app, rather than a nice flourish flair, and it comes with some usability downsides.
    a)  User arriving at this page -just- clicked that he wants to see some modlists.   Let's show him some modlists.  All that he sees is the tippy tops of some boxes way down at the bottom. And that's if they're lucky:
    b)  Some sizes of the window result in -no- mods showing on the page at all, with very little hint that they're down below.

**Are we on the start page here, or the browse modlists page?**
NVIDIA_Share_EnrU2cwHM7

eHCxjcTXe1

Overall, I think the bold WJ logo is nice, but it shouldn't be at the expense of not showing the controls people are there to use.  The layout should be designed in a way where it's a snazzy eyecatcher, not a blockade you have to swerve around to get to your goal.  I think we can shift things around to make it work.

### Create a Modlist 
Didn't notice it at first, but I do like the look of the blurred button when not active.  We should definitely find more ways to utilize that concept here and there.   I do think it needs a little more styling to make it clear that it's a "not selected" state.  Maybe gray it out a bit, too.  It looks like a low rez stretched image with just blur and no other effects, I think.

Overall though, I'm not a fan of this page.  
I think there's way too much empty space after having come from the massive WJ logo.   I think we can just go to the compiler view, and have just a combobox.  Have the inputs change with the combobox.

NyY96qOK4M

### Install Modlist From a File
Even more egregious than Create a Modlist as far as wasted and jarring empty space.  If someone showed this to me with no context, I would've put money this was a ported mobile app that had zero customization work done for desktop:

NVIDIA_Share_iOp7ZfIwkx

Once the modlist is selected, it's a -little- better.  But overall, I can never actually see everything at once.  I can see the title, or I can see the picture at blaringly large size, or I can see the input.  

aX3xmemLSN

### Installing Page
Mostly the same.  My problems with it are the two things that changed:
1) You added a second Top Bar to the existing Title/Progress Bar combo that was there.  Combo-ed with the implicit window bar.. we now have -three- mostly empty bars that take up a literal inch on my screen. 

(Side note:  Progress bar obviously not working until some backend is improved.  Also, just noticed a broken binding not displaying the modlist next to "installing" which I fixed on my branch yesterday).  

I had been aiming to eliminate the implicit window bar completely and customize the top bar.. but I'd be doing the opposite of this and collapsing it all into one bar.   I don't mind the added link logos, but they need to be put in the empty space we already have.

2)  You added a margin to my margin.  We already had a margin going around the edge (with the exception of the progress bar, which was an explicit design decision).   It's now wrapped like an amazon box in an amazon box.

NVIDIA_Share_HkdTYHnRF9

### TLDR
I think you went heavily out of scope of what a single PR should be, and made some major decisions while avoiding discussion.  I know we've had some pretty hefty ones before, but that's perhaps a side effect of being so early in the project.  I think we need to be more focused on one thing at a time, one PR at a time that's easy to digest.  
Small items like
  - Modlist Browser View extracted from front page
  - Made a styling adjustment to buttons to look different
  - Added back buttons

Two reasons:
1)  It helps more people work on things in parallel.  I've had trouble finding areas to put time into, because I know you've been absorbing more and more into this monolith PR.  The less things in the air at once, the more people can check out the current setup, make an improvement, and fold it back in.
2)  It helps @halgari merge the things that are good immediately.  If a PR is bite sized and scoped to one improvement, he can just glance over it and hit accept pretty easily.  With a massive PR, even if he likes all the items in a PR, he has to consider all of them right then.  Worse, if he doesn't like some of them, now that's a delay on accepting the part(s) he wants while the other parts are fixed up, even if they're unrelated they're now bound together.

I think you should extract your Mod Browser view to its own PR and submit that.  It shouldn't require any other changes besides a new ModListBrowserView and VM.  This way @halgari can grab it for the 1.0 release I think he wants to make soon without having to think too hard about all the ramifications.  I've got some issues with it I mentioned above, but we can discuss and do what we decide on as following PRs.

I think the other items need a lot more group discussion

halgari:
I agree, before @Noggog wrote this up, my suggestion was going to be to extract the first two parts of this PR and put them into their own PRs, so at this point what I'd love to see is:

1) the modlist browser pulled into its own PR. Even if this requires spinning up a new window, we can always integrate that window as a view into a single window later down the road
2) if we've pulled the browser out of the first page, then we'll need to re-design the front page. So the work you've done on the mode selection window (browse, install, compile) could be a second PR.

Wabbajack/View Models/CompilerConfigVM.cs
erri120: The ModManager enum is inside my `vortex-support` PR, will make sense once this is merged and I continue with `vortex-support`. 
I did try to use triggers and effects inside the XAML and failed miserably but can try again.

Wabbajack/View Models/CompilerVM.cs
erri120: Since the `MainWindowVM` now is just the main window which is parent to all other views, I thought it would be better to have elements only in VMs that actually use these:
The only VMs that use Work Queue and the Log are `InstallerVM` and `CompilerVM`, 2 out of like 10?. This is ofc under the assumption that the user doesnt try to compile and install in one WJ session, (havent tried that since we dont have a back button in either Installer or Compiler views)

erri120:
Big black buttons on front page:
I had no idea how to style them so they are a placeholder for whatever we want to use. 
Why I have ScrollView on nearly every page:
Different Resolutions have different amount of screen real estate we can occupy. If you maximise the WJ window on a 1920x1080 window, the banner will occupy the entire screen and you can start scrolling down to see the next steps. This is the same as on the website.
The problem with the startup page is that we dont really have stuff to display:
I thought about adding some text, like an introduction so it doesn't look empty but text in WPF is just god awful to add.
I originally had the entire gallery on the startup page but than I thought back at bombarding the user with stuff they may not even want/need to see.
There is a ton of empty space at the compiler config view since I wanted to move all settings from the compiler view to this config view. For MO2 you just need the profile but vortex will require a lot more than that.
Same for the installation config view:
You select the modlist and than depending on the type of mod manager used (mo2/vortex) different settings will populate the window. I honestly have no idea how else you would do this. The screen is empty with only one file picker because you only need one file picker and a text telling you what to do at that exact step. If you were to already show both MO2 **and** Vortex settings than you'd have like 30 settings on the screen and have no idea what to do. This whay it may look empty but it provides a guided and somewhat fluid experience to get to your goal: installing a modlist.
I will change the modlist image so the layout is similar to the current compiler view where you have the picture on the left and than properties on the right.

Wabbajack/View Models/CompilerVM.cs
Noggog: Yeah, you bring up a good point.  I think theres two halves to consider:
1)  The wiring of -how- to construct the list of CPU states and the list of Log events.  Rx is nice because we can construct a very complex observable that is instructions for how to get X, but we don't have to subscribe to it until we actually want that wiring to be set up and start running.
2)  The usage of that wiring to coalesce the results into a list for GUI display.


So your solution proposed in this PR is good for (2), as only the InstallationVM and CompilationVM would have any concepts of Log/Cpu setup at all.  If the user just went to browse, no CPU/Log compilation work will be done.  Only if they open InstallationVM/CompilationVM would that logic exist and take CPU.

Problem is now we're duplicating code in every VM that has interest in CPU/Log, so it's terrible for (1).

A good balance would be to define just the observable in a common place, either a static function, or as a member in a singleton class (MainWindowVM).   Then any interested VMs can reference that observable, subscribe, and coalesce into a list for display.  Best of both worlds.

Some caveats on that as well though.  Now if we do add back/forward buttons, if the user navigates to compiler and installer in the same run (not as rare as you might think for people devving), then we now have to sets of coalesce work being done.  This would get worse as more and more VMs are interested in CPU/Logs.

So one way to avoid this, is to not keep references to InstallerVM/CompilerVM as static Lazy members.   If the user navigates from installer to compiler, a compilerVM will be instantiated and set as the active pane.  The installerVM will be destroyed and cleaned up by the GC.  This means we're good to go again and we only ever have one interested party doing coalesce work.

Downside of that though is related to responsiveness.  Yeah, we're now doing less work overall, technically.. but the "wind up" time is heavier.  Things are not prepped, so when the user clicks between compiler/installer, the installer systems can't show until all the wiring has been constructed again, and all the logs coalesced again.  While probably fine, it starts adding up, and overall makes the app feel sluggish.

So I think it's a balance between doing as little work as you can, while also having most systems prepped and ready to go as quick as possible when the user clicks.  I think what we have already is a good balance.  The work is being done when some parties aren't interested, but at least it's being done only once, and has the upside of being ready immediately when any parties are involved, and the other upside of not having to be reconstructed when transferring between two interested parties.

Noggog:
### Big Black Buttons (BBB)
Yeah, I think Tiles in the first place are just buttons /w a specific style to look like windows phone tiles.  The actual execution of that by mahapps seems to be shit, so they're basically just buttons with no style.  If we want to improve them, we should just go back to normal buttons design a style we like with a black background as a starting point (as that's all the tiles seem to be, lol).

>  If you maximise the WJ window on a 1920x1080 window, the banner will occupy the entire screen and you can start scrolling down to see the next steps
I don't view this as a positive.  We're playing peekaboo with the WJ face on -every- section navigation.

> There is a ton of empty space at the compiler config view since I wanted to move all settings from the compiler view to this config view. For MO2 you just need the profile but vortex will require a lot more than that.
I think what you're pitching here is a drastic re-think of the foundations of the app's presentation mentality.  You're pitching that every screen the app ever displays will be one scrollviewer /w controls showing line by line.  Controls pop up at the bottom as the become relevant, like an instagram feed.  I think this is a philosophy and major design shift from our current state, and at bare minimum, @halgari  would need to pass off that this is the design route he wants to go down.  Until we've all had the discussion and come to an understanding, I wouldn't spend all your hours hammering it out.  I'd highly recommend bringing it up in chat and discussing the pros/cons for all of us to consider first.

PR #158, opened 2019-11-09
Created new ModList gallery

erri120:
Unsure about the new downloader


Noggog:
Cool, looking like a good start.  I think you can also bring over and/or add the swap in of this view in place of the old modlist gallery.  Old setup is still 100% present and being used.

  - Need a VM to power this, which can be swapped into ActivePane to show the view here like installer/compiler do.
  - If you want to, I think you can do some surgery on the ActivePane property while we're here.  I don't actually like the enum powering the pane swapping.  I think it can be just a plain Reactive property that people set normally.  Commands would literally just do `MVWM.ActivePane = theModlistGalleryVM;`.  Eventually the RunMode enum can be nuked, imo... but that might require other refactors we can punt on and do in a later PR.
  - Once the VM is ported over or added, all the OnClick codebehind mechanics should be swapped for bindings to ICommands.
  - I think the choices in how to handle the downloader don't matter too much, as I expect we'll be polishing it separately pretty soon.  Whatever gets it somewhat functional and like the existing setup should be okay.

PR #159, opened 2019-11-09
Better Game Handlers

erri120:
Support for:
- GOG
- Steam


halgari:
Looks great, it's purely additive (doesn't modify existing code) so simple enough to review.

PR #180, opened 2019-11-16
Vortex stuff

erri120:
- abstract class for Vortex and MO2 installer
- moved core installation functions to AInstaller so we don't have to maintain two Installer that use the exact same functions
- moved core compiler functions to ACompiler
- updated the VortexCompiler with more quality of life changes


erri120:
now mergable

PR #182, opened 2019-11-17
Instanced Work Queues

halgari:
WorkQueues are now instances created by the compilers/installers, and passed to the VFS. Every call to PMap now takes another parameter of a WorkQueue on which to dispatch jobs. @Noggog I had to comment out the old Rx code that updates the UI for the work queue, because there is no longer a global updater, instead we need to attach the UI to the Compiler/Installer `Queue` objects. Shouldn't be too hard, but I'm not really sure the best place to wire that up in the UI, so I thought I'd let you take a look. Everything seems to be working fine, it's the "CPU status" section of the UI no longer gets updates. 

Making work queues non-global will allow us to create the queue at compiler execution time, which will allow us to examine the work that's about to be done and adjust the size accordingly. Perhaps we can even add a "low priority" button on the UI that would halve the number of threads created for those people who want to use their computer during a WJ install.


Noggog:
Sounds like a good change.  Could even eventually make it dynamic to shrink/grow mid-install if the user decides it's too heavy halfway through.

I'll branch off this and adjust the UI

Noggog:
Made a PR to pull my changes into this PR?  Not sure if that was the proper way to do it

PR #187, opened 2019-11-18
Nitpicking stuff #1

erri120:
ModList list from GitHub will display the description of a game instead of toString if available

### DIFF
--- a/Wabbajack.Lib/ModListRegistry/ModListMetadata.cs
+++ b/Wabbajack.Lib/ModListRegistry/ModListMetadata.cs
@@ -6,6 +6,7 @@ using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
 using System.Windows.Media.Imaging;
+using CommonMark.Syntax;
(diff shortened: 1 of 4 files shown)


Wabbajack.Lib/ModListRegistry/ModListMetadata.cs
```diff
@@ -31,6 +32,9 @@ public class ModlistMetadata
         [JsonProperty("game")]
         public Game Game { get; set; }
 
+        [JsonIgnore]
```
Noggog: I think this should just be a getter property like
`public string GameName => this.Game.ToDescriptionString();`

Rather than an actual field we have to set, that could be potentially desynced (Skyrim `Game` enum, /w `GameName` set to "Fallout")

Noggog: If you want the fancy functionality of using description if it's set, otherwise fall back to `ToString` I'd just define and extension function to do that logic `DescriptionOrToString`

PR #191, opened 2019-11-19
updated intro of readme for clarity

ichikuma7:


halgari:
These don't seem to be much of an improvement, and in some cases the changes don't make sense at all. Sorry, I'm going to have reject this PR.

PR #196, opened 2019-11-20
Disk size and performance

halgari:
Set the installation queue size based on the disk speed and not just the CPU.

* we benchmark every folder involved in the installation (downloads, scratch space, and installation)
* find the slowest speed of each
* if the speed is < 100MB/sec we use 2 threads, for each 50MB/sec above that we get another thread to a max of the number of CPU threads

We also now calculate required space and throw an error if there's not enough available.


erri120:
Add the test result to the settings file so we dont do this all the time

halgari:
It only runs for 2sec, but it does occur to me that we can cache the results, and reuse them for each drive. I'll implement that before merging this PR.

PR #198, opened 2019-11-21
Cleanup

erri120:
The readmes for each project will be shown when looking at the code for each project on GitHub. ([Example](https://github.com/erri120/wabbajack/tree/nitpicking-3/Wabbajack.Common.CSP))
I removed the old VFS project since we now have `Wabbajack.VirtualFileSystem`.
I want to clean up a bit and started with removing all unused imports. This lead to the removal of around 600 unused imports. I split the commit for removing unused imports into one for each project and one for all tests so that the commit history does not look completely _fucked_.
Also simplified naming by removing `this` and fixed some variable names:
`private int _Something` => `private int _something`
`public int something` => `public int Something`
`_something_another` => `_somethingAnother`


Wabbajack.Lib/ModListRegistry/ModListMetadata.cs
```diff
@@ -1,18 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
```
Noggog: I thought resharper wasn't removing these? 8P

erri120: Lol, apparently flips a coin each time it sees it. Some still have it

Noggog: Not a huge deal since it's already done, but if there's a setting somewhere deep in there to tweak for future cleans might be nice

Noggog:
Renaming looks good to me overall.  Can't speak on the merged projects as I'm not familiar with them.

PR #208, opened 2019-11-24
Progress Bar Systems Rewired

Noggog:
Both compiler/installer sides should display progress now.

They seem to just go from 0 to 100 with no progress, but assumedly that's just because the backend needs to add that reporting granularity.

I will mention that I noticed the progress systems seem to have several phases that reset progress to zero.  `StatusUpdateTracker.NextStep` specifically.

If we want to continuously reset progress to zero in this fashion, then I think we need to add another layer to the progress reporting.

- Current Step
- Amount of Steps Total
- Current Progress of Current Step (this is currently what's being communicated)

Otherwise the progress bar displayed to the user is just going to repeatedly hit 100% and then drop to 0%.  If we can know the current step and max steps, then we can do some logic to display it appropriately.

### DIFF
--- a/Wabbajack.Common/StatusUpdate.cs
+++ b/Wabbajack.Common/StatusUpdate.cs
@@ -38,9 +38,9 @@ namespace Wabbajack.Common
             _progress.OnNext(0.0f);
         }
 
-        public void MakeUpdate(double progress)
+        public void MakeUpdate(float progress)
         {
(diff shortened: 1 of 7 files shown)


Noggog:
Hold off a bit on merging this.  Looking into one oddity I'm seeing

Noggog:
The issue I'm seeing happens on the current master, so I think this PR is good to go.

PR #214, opened 2019-11-26
Add Steam Workshop downloader (don't merge yet)

tr4wzified:
Still need to optimize a lot and broken atm with hardcoded Skyrim LE game id. Just want to put it out here. Also need to use SteamHandler


Wabbajack.Lib/Downloaders/SteamDownloader.cs
```diff
@@ -0,0 +1,58 @@
...
+using Wabbajack.Common;
+
+namespace Wabbajack.Lib.Downloaders
+{
+    public class SteamDownloader : IDownloader, IUrlDownloader
+    {
+
+        public AbstractDownloadState GetDownloaderState(dynamic archiveINI)
+        {
+            var url = archiveINI?.General?.directURL;
```
erri120: Since workshop mods do not have an ini file like MO2 mods or vortex downloads, after they go through the pre compilation steps in the vortex compiler, I'd suggest looking at the pre compilation steps from the vortex compiler for creating a new ini file for these downloads.
steamworkshop inis would have a new field `isSteamWorkshop` or something like that so we can know what Downloader to use.

PR #216, opened 2019-11-27
Port to .NET Core 3

halgari:


halgari:
I think we'll let this one sit till after V1. As much as I'd like to switch to .NET core the changes in this PR are too fundamental for us to properly QA test before release. We'll keep this PR though and I can dive back into it after V1

PR #223, opened 2019-12-01
Extract .exe with innounp

erri120:


halgari:
Would be great to get a test for this: download a .exe from the nexus and check the extraction. The BSA tests do this, so you could copy/generalize some of the caching code from those tests.

PR #225, opened 2019-12-02
integrate modals into the main window

halgari:


halgari:
performed as part of another PR

PR #226, opened 2019-12-02
Some Fixes and Error Handling

Noggog:
- fix for HTTPDownloader if the target path's directories didn't exist
- Cleaned up startup code (migrated all to mainwindow ctor).  Fixed window size defaults not being applied
- Some error handling for if installer is pointed to a corrupted modlist, rather than crashing


Noggog:
Going to rebase and reopen

PR #236, opened 2019-12-05
Exceptions as data thoughts

Noggog:
Decided to do my thoughts as a PR, just because it'll be easier to communicate the ideas with granular commits, I think.  Don't feel like you gotta merge all of them. Feel free to just take any keeper thoughts and cherry pick or reimplement.

Overall definitely what I was thinking of for the system we should use.  I think we'll be able to do some neat stuff with this.

You asked specifically about my thoughts on ShortDescription / ExtendedDescription.  I do like the concept a lot.  Short can be used for display in the log to keep things neat, and Extended can be exposed via tooltip, or a click-to-expand setup.  Awesome

Other Thoughts Expressed via Commit:
### IException, GenericException
When funneling every log message through an Rx pipe, there wasn't an easy way to take an Exception in a catch block, and put it into the logging system.  You'd have to convert it to a string right then, which means our logging system loses access to the Exception object if it would find it useful.  This interface should allow us to quickly put any exception down the pipe, not just our custom error classes.

### Made Utils.LogToFile private.  Removed ExceptionToString
LogToFile was exposing a route to the user that would log a message to the file without going to the Rx pipe.  Closed that off.

Added `Utils.Error(Exception ex, string extraMessage = null)`

Removed ExceptionToString, as users should just use the above method, which will let us change how we print exceptions, or do extra things with them easily in the future if we want to.  I think it's good practice to not lose the Exception object by converting it to a string early before it hits the centralized logging systems.

Went around and cleaned up all the old exception printing messages that broke.

### Utils.ErrorThrow added for when wanting to throw the exception
`Util` now offers two functions for errors: `Error` and `ErrorThrow`.

This is mainly to provide symmetry in API when considering the `Util.Error(Exception ex)` call.

For example, consider a user doing this:
```
catch (Exception ex)
{
    Util.Error(ex);
}
```

Does this throw the exception?  It shouldn't because the user needs the option to report an error but not throw.  So `Error` should be just reporting the error, and the user can throw themselves afterwards (or call `ErrorThrow`).

But now `Util.Error(Exception ex)` doesn't throw, but `Util.Error(AErrorMessage err)` does, which will definitely lead to confusion and bugs having that asymmetry.   Solution seemed to make the throwing version obvious in the name `ErrorThrow` and have a `Util.Error(AErrorMessage err)` to be symmetrical even if it's mostly unused.

Edit:
On thinking about this more, I think it would also be fine to just rename `Util.Error(Exception ex)` to just `Util.Log(Exception ex)`.  That way you can leave `Util.Error(AErrorMessage)` as the throwing call if that's a more preferable name.  I still lean the `ErrorThrow` way, though, as I could see myself accidentally calling `Error` without realizing it would also throw internally.

### Made lots of error classes members public gets
I think one of the real selling points of making messages/errors into classes is that now the GUI (or whoever) can look for specific message types, and offer customized behavior for each when it wants to.  A big enabler of that is having access to gritty members, so it can place member A here as red, and B there as green, and show C on hovering, etc.  If everything is coalesced to a single string for it to use, it makes that way more difficult.

So, the members have been made public getters, so the GUI can access them later on, rather than solely relying on Short/Long message, they can go more granular if it's helpful.

Short/Long message would be the fallback for when the GUI doesn't have customized behavior for that class, and also will be used as it falls through to the Util.LogToFile methods for printing to disk, so they'll still be important.

### Removed IUserIntervention.Resume
One thing to consider, is that every user intervention fired down the pipe will probably be explicitly recognized, cast to its type, and handled individually based on what Type it is.  I don't foresee getting a `IUserIntervention<string>` and just handling it in its interface form.  If I get a `IUserIntervention<string>`, I have no clue if I'm opening a file picker, or just offering a textbox, or maybe getting the string some other way?   The GUI's clue for what the engine wants is the Type of class that it is (MissingFileIntervention, open FilePicker), so it must recognize and cast the intervention class to proceed.

Well now that we've cast it, and my code is dealing with the literal "MissingFileIntervention" class, we don't have to limit ourselves in the Resume definitions.  It could be `Resume(string path, string nickname)` or `Resume(int maxCpuCores, int maxMemoryUsageMB)` or whatever that individual intervention needs.  I think the `IUserIntervention<T>.Resume(T item)` holds the system back from having more expressive communication.  We have lots of communication going from the engine to the GUI in the form of members exposed, and this would allow just as vibrant communication the other direction.

I think `IUserIntervention.Cancel()` is still important.  Obviously it'll be called if the user clicks "cancel" or can't find the missing file, or whatever... but it's also the fallback for us.  If the GUI looked through all its known interventions and couldn't cast the object, well then it can't handle it.  It then calls `Cancel()` to signify it's got no clue what to do.  This would only ever happen if we messed up, as every intervention should be known by the GUI-side, too.  So we'd probably see this in user reports and fix the mismatch up.

I do think an argument could be made for `IUserIntervention<T>.Resume(T item)` as a way to implement a generic handler system if we felt like we couldn't/wouldn't implement every intervention explicitly on the GUI side.  The GUI could use reflection, find the type of T, and if `string`, make a very standard display with the ExtendedDescription + a TextBox, or an IntSpinner if it was an `int`.   However, even then, I'd say my preferred setup for handling this requirement would just be to add a `GenericStringIntervention` class.

### WorkQueue.Messages removed (for now)
I did this mainly in the short term until we sort the plan out.  It was a bit confusing to have message subjects on each WorkQueue instance, while also having a static one in Utils.  If I'm the InstallerVM, and I've just kicked up a new install which has its own WorkQueue, do I hook onto the WorkQueue's subject or onto Utils?  It wasn't exactly clear in its current state.

I think if we opt to have separate message subjects in each WorkQueue and one in Util, we'll want to do the following first:
1)  Always funnel WorkQueue's messages into Util, as it does the file logging.  We want everything logged (I assume) so it'll have to get there.
2)  So now Util has all the messages of everything, while WorkQueue's should have all the messages related to its run?  But that might not be true, because subfunctions a WorkQueue is running might go call the static Util's logging methods, meaning they wouldn't fire on the WorkQueue's subject they were working under.  To really be reliable we would want to change all calls to fire through their active WorkQueue's subject, and only calls truly unrelated to any WorkQueue call Util.

But that seems like a bunch of work that probably doesn't matter, and right now having two subjects wasn't reliable of which logs went where.. so I just simplified it to the singular Utils for now.  I do think it would be more ideal to have a message stream for each WorkQueue (in case we have two running at once?), but I think we'd have to do a lot more work to prep that.

### DIFF
--- a/Wabbajack.Common/FileExtractor.cs
+++ b/Wabbajack.Common/FileExtractor.cs
@@ -48,12 +48,11 @@ namespace Wabbajack.Common
                 else if (source.EndsWith(".exe"))
                     ExtractAllWithInno(source, dest);
                 else
-                    ExtractAllWith7Zip(queue, source, dest);
+                    ExtractAllWith7Zip(source, dest);
             }
             catch (Exception ex)
             {
-                queue.Log($"Error while extracting {source}");
-                throw ex;
+                Utils.ErrorThrow(ex, $"Error while extracting {source}");
             }
         }
 
@@ -154,14 +153,13 @@ namespace Wabbajack.Common
             }
             catch (Exception ex)
             {
-                queue.Log($"While Extracting {source}");
-                throw ex;
+                Utils.ErrorThrow(ex, $"While Extracting {source}");
             }
         }
 
-        private static void ExtractAllWith7Zip(WorkQueue queue, string source, string dest)
+        private static void ExtractAllWith7Zip(string source, string dest)
         {
-            queue.Log(new GenericInfo($"Extracting {Path.GetFileName(source)}", $"The contents of {source} are being extracted to {dest} using 7zip.exe"));
+            Utils.Log(new GenericInfo($"Extracting {Path.GetFileName(source)}", $"The contents of {source} are being extracted to {dest} using 7zip.exe"));
 
             var info = new ProcessStartInfo
             {
@@ -210,7 +208,7 @@ namespace Wabbajack.Common
             {
                 return;
             }
-            queue.Log(new _7zipReturnError(p.ExitCode, source, dest, p.StandardOutput.ReadToEnd()));
+            Utils.Log(new _7zipReturnError(p.ExitCode, source, dest, p.StandardOutput.ReadToEnd()));
         }
 
         /// <summary>
--- a/Wabbajack.Common/FileExtractor.cs
+++ b/Wabbajack.Common/FileExtractor.cs
@@ -83,7 +83,7 @@ namespace Wabbajack.Common
             }
             catch (Exception e)
             {
-                Utils.LogToFile($"Error while setting process priority level for innounp.exe\n{e}");
+                Utils.Error(e, "Error while setting process priority level for innounp.exe");
             }
 
             var name = Path.GetFileName(source);
@@ -104,7 +104,7 @@ namespace Wabbajack.Common
             }
             catch (Exception e)
             {
-                Utils.LogToFile($"Error while reading StandardOutput for innounp.exe\n{e}");
+                Utils.Error(e, "Error while reading StandardOutput for innounp.exe");
             }
 
             p.WaitForExit();
--- a/Wabbajack.Common/StatusFeed/AErrorMessage.cs
+++ b/Wabbajack.Common/StatusFeed/AErrorMessage.cs
@@ -6,10 +6,11 @@ using System.Threading.Tasks;
 
 namespace Wabbajack.Common.StatusFeed
 {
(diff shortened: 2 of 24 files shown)


Noggog:
Oh, you mentioned 
>  it would be awesome if the tooltips for the ExtendedDescription had links to offending files, links to mods on the Nexus, etc.

I totally agree.  I think the public getter API change is the way to go, exposing `NexusLink` as a member, vs embedding link strings in ExtendedDescription.  It'll give us more flexibility in how we choose to display things.

PR #241, opened 2019-12-07
Async exploration

Noggog:
- WorkQueue now takes in async jobs
- PMap returns task
- Thread aborts swapped for cancellation tokens
   This is a bit more verbose, especially in the ABatchProcessor subclass code, as anywhere we might potentially want to exit a job early, we check cancellation status and return.  We could look into reversing this change and going back to abort, but I think it might be problematic in a few situations.
- all Result calls swapped for awaits

I wouldn't merge this too quickly, as it is a pretty major refactor.  Definitely check it out and run it yourself and maybe look around a bit before considering a merge.

One thing that was concerning while refactoring was that any function returning `void` did not trigger the compiler error if it made calls and didn't await their tasks.  This I suppose is because the compiler sees that function as an `async void` function, which is technically valid.  This has the unfortunate side effect of not allowing us to 100% use the compiler to tell us if there's any missing awaits. I attempted to be very rigorous during the refactor, and followed references after changing every function to make sure all its callers did the await, but I very well could have missed a few.


Wabbajack.CacheServer/NexusCacheModule.cs
```diff
@@ -18,6 +18,9 @@ public class NexusCacheModule : NancyModule
 
         public NexusCacheModule() : base("/")
         {
+            // ToDo
+            // Handle what to do with the fact that lots of these are now a tasks
+            throw new NotImplementedException("Unsure if following functions still work when taking in a Task");
```
Noggog: This area might be of interest to look at before merging.  This code didn't seem to be used, so I left it in a non-functional state, as I wasn't quite familiar with the NancyModule to know its goals or how to handle the new async return values.

Wabbajack.Lib/Downloaders/NexusDownloader.cs
```diff
@@ -75,12 +77,13 @@ public override bool IsWhitelisted(ServerWhitelist whitelist)
                 return true;
             }
 
-            public override void Download(Archive a, string destination)
+            public override async Task Download(Archive a, string destination)
```
Noggog: Code refactors like this are where one potential danger lies.  A previously `void` function that now is `async Task` is the proper swap, but the compiler will not reliably complain if it hadn't been changed and was left void.  So the potential exists that some other function out there needs to be `async Task` but is still void.

Wabbajack.Lib/MO2Compiler.cs
```diff
@@ -112,10 +116,12 @@ protected override bool _Begin()
                         { Path = Path.Combine(Consts.LOOTFolderFilesDir, p.RelativeTo(lootPath)) });
             }
 
+            if (cancel.IsCancellationRequested) return false;
             UpdateTracker.NextStep("Cleaning output folder");
             if (Directory.Exists(ModListOutputFolder))
                 Utils.DeleteDirectory(ModListOutputFolder);
 
+            if (cancel.IsCancellationRequested) return false;
             UpdateTracker.NextStep("Finding Install Files");
```
Noggog: This verbosity is a side effect of swapping the systems to CancellationTokens instead of thread aborts.   It's more compatible, as code that does `Task.Run()` calls cannot use the abort.  These would be seen in places like MO2InstallerVM's BeginCommand.   Eventually we'll want to add "stop installation" functionality, where we'd pass a cancellation token into the MO2Installer.Begin() call.

So I feel like this cancellation checking verbosity is just a necessary downside. 

Additionally, I think we need to extend it further, by passing the cancellation token into heavier steps, to have those steps check it periodically so they can also exit out early when appropriate.  Otherwise if cancellation is requested on a heavy step that takes 10m, it won't actually cancel for those 10 minutes if that step doesn't ever check the token.

Wabbajack.Lib/MO2Installer.cs
```diff
@@ -178,27 +194,28 @@ private void AskToEndorse()
...
             Info("Done! You may now exit the application!");
         }
 
-        private void BuildBSAs()
+        private async Task BuildBSAs()
         {
             var bsas = ModList.Directives.OfType<CreateBSA>().ToList();
             Info($"Building {bsas.Count} bsa files");
 
-            bsas.Do(bsa =>
```
Noggog: Could potentially be refactored to a Task.WhenAll if we want these to be done in parallel.  Not sure if that's desired or not, so I left it serial.

Wabbajack.Lib/NexusApi/NexusApi.cs
```diff
@@ -50,15 +51,19 @@ public UserStatus UserStatus
...
 
-        public bool IsPremium => IsAuthenticated && UserStatus.is_premium;
-
-        public string Username => UserStatus?.name;
+        public async Task<bool> IsPremium()
+        {
+            return IsAuthenticated && (await UserStatus).is_premium;
+        }
 
+        public async Task<string> Username() => (await UserStatus).name;
```
Noggog: The NexusApi required some discretionary choices, as the ctor and these properties were doing async stuff.  The constructor definitely needs to be a factory instead, but we could opt to offer these properties without tasks, if we want to do them at the time of the factory, rather than lazy retrieval upon access.

Noggog:
I think the above comments were all my concerns.  If those all look good, then I'm good with this PR on my end.

PR #255, opened 2019-12-11
Intervention frontend

Noggog:
- GUI displays added for MO2 installation replacement intervention query.
   - Has a "remember next time" option
- Added `CpuID` and `bool Handled` to IUserIntervention
- Made AUserIntervention, which all user interventions inherit from
   - They implement ReactiveObject, and `Handled` raises property changed
     This is so that users can track if/when an intervention is handled, which helps display
- Readded LogMessages member on WorkQueue, which is "fake".  Decided to do this so that users could attach to WorkQueues with a lifetime to "get" their log messages.  It currently just routes to the singleton LogMessages stream to actually fulfill.  If we ever do scope the LogMessages to be instanced properly, then all the wiring is set up ready to go for that.
- ABatchProcessor exposes its WorkQueue's LogMessages


Noggog:
Also slide in the installation complete view I had from a while back.  Might adjust its look a bit sometime, but it should be fine for now

Noggog:
Ehh, just realized I gotta put the completion setup on the compiler side, too.  Merge if thou wilt

PR #256, opened 2019-12-12
Bugfix for case sensitivity in downloads folder detection

Noggog:
Small fix for cleanup logic being case sensitive when filtering for downloads.

Couldn't think of any quick good unit tests that wouldn't require refactoring the surrounding code a good bit.  Tested via breakpoints and tinkering /w path in WJ GUI.


halgari:
ack, I didn't see this before I coded up my solution. Sorry for the duplicated work

PR #266, opened 2019-12-13
Deadlystream support

erri120:


halgari:
You'll probably want to rework this against #353 once it gets merged.

erri120:
Will redo this

halgari:
@erri120 it looks like Deadlystream might also be a IPS4 site? If so look at the implementation of VectorPlexus as any IPS4 site should be implementable in about 20 lines of code

erri120:
yeah just saw it

PR #287, opened 2019-12-17
Made HashSets case insensitive

tr4wzified:
I saw two ba2's floating around in the same HashSet. Made it case insensitive, aswell as the others at the same time.


halgari:
Why did I never think of doing this! Thanks!

erri120:
> 
> 
> Why did I never think of doing this! Thanks!

557e44368553e2430730ad7d446d33c9058a032a

tr4wzified:
Yeah, I got inspired by erri :)

PR #291, opened 2019-12-18
Fix error on HTTPDownloader trying to download to non-existing destinations

tr4wzified:
A bunch of users had errors when downloading modlists, Wabbajack tried to download them but the destination didn't exist. Destination directory is now automatically created if it doesn't exist yet.


halgari:
If `download == false` then don't try to create the paths. `download == false` is used to tell the function that WJ wants to verify that the link is live, but not write anything to disk

tr4wzified:
Alright, will add

tr4wzified:
Done

PR #297, opened 2019-12-20
Bugs n features

Noggog:
- Fix for progress bar showing 100% initially
- Refactored Cef file extractions.  Was missing an await
   Also changed it to do the work in the background, while also adding safety by making web browser wait until it's complete before opening.
- Logic to move WJ to front of other apps when starting
- Install/Configuration failure display as red
- Fix for Modlist browser's "need download" check on modlists.  (was failing to read file to hash while modlist was downloading and locked)
- Readmes now open after installing, and after downloading a modlist from the browser
- Readmes can now be websites


Noggog:
Hmm, Don't merge this just yet.  I just realized my commit that added websites for readmes corrupted the modlists.

Noggog:
Ceras version tolerance enabled.  Modlists will still be corrupted if this is merged, but it'll be safe going forward to add new members

PR #301, opened 2019-12-21
Minor changes

tr4wzified:
- Fixed existential crisis in the logs, along with other typos
- Remove unnecessary reference to library Wabbajack.Common


Wabbajack.Lib/MO2Installer.cs
```diff
@@ -45,14 +45,13 @@ protected override async Task<bool> _Begin(CancellationToken cancel)
             ConfigureProcessor(18, await RecommendQueueSize());
             var game = ModList.GameType.MetaData();
 
-            if (GameFolder == null)
-                GameFolder = game.GameLocation();
```
Noggog: Don't think this is good.  The reason this has an if statement is that it allows the creator of a MO2Installer the ability to specify a GameFolder explicitly, if they desire.  The if statement here, then, makes it act like a fallback default for when it's not set.

ACompilerTest L61 is an example of this being utilized

tr4wzified: Missed that, woops. Reverting

tr4wzified:
Added the if back now.

PR #340, opened 2020-01-04
Added null check to two dimensional array

erri120:
My take on the ScreenSize problem. As I said on the Discord: `data.Sections` appears to be a two dimensional array and we only check if `data.Sections["Display"]["iSize W"] != null` but we did not check if the `Display` section exists in the first place.
Also added `CultureInfo.CurrentCulture` as the `ToString()` argument.

### DIFF
--- a/Wabbajack.Lib/MO2Installer.cs
+++ b/Wabbajack.Lib/MO2Installer.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Globalization;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


halgari:
Why do we need culture info for a integer?

erri120:
> 
> 
> Why do we need culture info for a integer?

ReShaper being picky: 
https://www.jetbrains.com/help/resharper/2019.3/SpecifyACultureInStringConversionExplicitly.html

PR #343, opened 2020-01-04
Compilers will include Reason when for no match in output

erri120:


halgari:
Nice! We should have this a lot earlier

PR #348, opened 2020-01-05
Downloader adjustments

Noggog:
PR with some odd one-off downloader improvements/fixes.   These are cherrypicked from a larger PR, so they may seem random.

No need to merge this immediately. Wanted to just start a PR on these so that I could point out a few things before I forgot about them, rather than waiting for a larger PR later.

- Nexus/LL downloader commands now try/catch their executions.   ReactiveCommands get really unhappy if exceptions are thrown (and stop firing new executions).
- New Utils.CatchAndLog convenience call.
- ViewModel/INotifyPropertyChanged concepts removed from INeedsLogin related classes.
- Global user intervention handling now try/catches so it can not halt if an error occurs.
- Downloader user intervention cancellation calls now call `TaskCompletionSource.TrySetCancelled`.  The `.Cancel()` call throws if you do it twice in a row, which was a bit obnoxious.  Would prefer to be able to cancel twice without it throwing.  (It does so currently in the wrap web browser logic, for ex)
- Removed some vestigial members (unrelated)

### DIFF
--- a/Wabbajack.Common/Utils.cs
+++ b/Wabbajack.Common/Utils.cs
@@ -129,6 +129,30 @@ namespace Wabbajack.Common
             }
         }
 
+        public static void CatchAndLog(Action a)
+        {
+            try
+            {
+                a();
+            }
+            catch (Exception ex)
+            {
+                Utils.Error(ex);
+            }
+        }
+
+        public static async Task CatchAndLog(Func<Task> f)
+        {
+            try
+            {
+                await f();
+            }
+            catch (Exception ex)
+            {
+                Utils.Error(ex);
+            }
+        }
+
         /// <summary>
         ///     MurMur3 hashes the file pointed to by this string
         /// </summary>
--- a/Wabbajack.Lib/Downloaders/INeedsLogin.cs
+++ b/Wabbajack.Lib/Downloaders/INeedsLogin.cs
@@ -8,7 +8,7 @@ using System.Windows.Input;
 
 namespace Wabbajack.Lib.Downloaders
 {
-    public interface INeedsLogin : INotifyPropertyChanged
+    public interface INeedsLogin
     {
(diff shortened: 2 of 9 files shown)


Noggog:
## ViewModel/INotifyPropertyChanged Removal
It's my opinion that both those concepts are unnecessary in classes that aren't meant to be in a GUI.  The INeedsDownload are -sort- of bound to by the GUI, as they do latch onto the `TriggerLogin` command.  But that's a very light interaction.

The ViewModel/ReactiveObject concepts are used in WPF GUIs because they bind using INotifyPropertyChanged.. and can't "bind" to an Rx observable/subject.  The ViewModel is sort of the hacky glue marriage between Rx and WPF, and should really only be used in scenarios where Rx observables need to be "manifested" into plain ol' members so that WPF can bind to them (usually via the OAPH concepts that get added by the [Reactive] tag for us).

TLDR:  In "engine" classes, we don't need `ViewModel`.  We can just expose members we want to be "subscribable" as either subjects or observables, depending on if we want them readonly.  INeedsLogin already did so with its `IsLoggedIn` member, so ViewModel inheritance was doing nothing for us anyway currently.

Noggog:
## TaskCompletionSource cancellations within our UserInterventions
I think we should just establish the pattern that we do `TrySetCancelled` rather than `Cancel` in our UserInterventions.   Otherwise we'll get "Task was already cancelled" exceptions every so often, which really adds nothing for us in our UserIntervetions system.  We should be fine with cancelling an intervention multiple times.

I just bring it up so we can mentally note it for any future UserInterventions we write

halgari:
These are all solid improvements, no reason to hold back on the merge.

Noggog:
Yeah, was more a vehicle to have the discussion and bring up the notable points, haha

PR #350, opened 2020-01-05
Navigation Stack

Noggog:
This is part of a pair of PRs meant to be a choice between alternative navigation patterns.  The other choice is (https://github.com/wabbajack-tools/wabbajack/pull/349).  One or the other should be accepted.

This PR goes the route of utilizing a navigation stack, and fully embraces it.  MainWindowVM is made "smart" and utilized a navigation stack member to put VMs on, and pop them off when appropriate.  VMs are made more dumb, and have less knowledge of what the stack's situation is, they just tell it to put something else on, or pop something else off.

- `MainWindowVM.ActivePane` is made private.  Navigation is done via NavigateTo() and NavigateBack() calls.
- A basic settings panel is added, which highlights some of the challenges in navigation.

Pros:
  - Specific VMs need less context information to function, so can be a little simpler.  LoginManagerVM doesn't need to know who opened it, for example.  It just calls NavigateBack().

Cons:
  - There are a few traps this pattern can fall into:
     - Insufficient/Unclear use of NavigateBack().  VMs can just repeatedly call NavigateTo() between two alternative views and technically "work"... rather than popping the current one off and putting themselves on, when appropriate.  This could lead to a stack that's hundreds of VMs deep:  `InstallerVM -> MainViewVM ->InstallerVM -> MainViewVM ->InstallerVM -> MainViewVM ->etc.`   It's sometimes unclear when NavigateTo should be called, vs popping first and then using NavigateTo.  
     - NavigateBack() has to be called carefully, and only once when appropriate.  Currently this PR suffers from this misuse, where the web browser logic actually calls navigate back twice, leading to unintended behavior.  This is highlighted by this video, where you can see the backwards navigation from the login page hops over the settings page, and goes back to the main selection view.  It should be going back to the settings page, if navigate back was called once, as appropriate.

Just some thoughts to consider.  Can still be a workable pattern, just highlighting the typical edge cases we'd have to deal with.

Link to example of NavigateBack misuse:
https://discordapp.com/channels/605449136870916175/620954238931173378/663224008589705236


Wabbajack/View Models/UserInterventionHandlers.cs
```diff
@@ -27,13 +26,12 @@ public UserInterventionHandlers(MainWindowVM mvm)
         private async Task WrapBrowserJob(IUserIntervention intervention, Func<WebBrowserVM, CancellationTokenSource, Task> toDo)
         {
             CancellationTokenSource cancel = new CancellationTokenSource();
```
Noggog: This is the code that's misusing NavigateBack, for reference.  Obviously this can be bugfixed, but I left it in to highlight how easily it can be misused.

PR #361, opened 2020-01-06
Deadlystream Support + VectorPlexus fixes

erri120:


halgari:
A test for the downloader would be great

halgari:
Needs a test for the DeadlyStream code, but it should work, so I'll merge it for now so I can get a release out the door.

PR #381, opened 2020-01-11
Cores Usage Settings

Noggog:
A more experimental PR that adds the ability for the user to limit the amount of active Tasks pulling work from the WorkQueue. 

Since this PR modifies WorkQueue logic, it would be good to have a few sets of eyes look at the changes to confirm correctness.

### WorkQueue Refactors
- WorkQueue swapped from using Threads to using Task.Run() calls.
- WorkQueue constructor now takes `IObservable<int> numDesiredThreads` rather than `int`.
  As the number of desired threads changes, WorkQueue will add new Task.Run() calls that pull work from the queue.  As it decreases to lower values, Tasks will shut down and remove themselves until the desired number of workers is active.  This removal only happens after a queue job has been completed, so might not happen immediately.
- Queue take system swapped to TryTake.  This allowed for more responsive shutdown of idle tasks when desired threads is lowered.
- Unit tests for various WorkQueue functionality

### UI/Settings Panel Refactors
- Some more experimentation /w RxUI "binding" systems, rather than using XAML bindings.  Seems to be going well enough.  The whole app is still a mix of the two, though, as they can be interchanged.
- Settings panel is in a "complete" state, but might evolve as time passes.  I was going to do a "tab control" based settings panel, but it looked comically empty with the limited settings we have.  So I just mashed everything into a wrap panel for the short term until we get more settings overall.
- Added frontend "Usage Percent" slider in the CPU view, so if users can modify the % as the program is running without going to the settings panel.


Wabbajack.Lib/ABatchProcessor.cs
```diff
@@ -12,7 +13,7 @@ namespace Wabbajack.Lib
 {
     public abstract class ABatchProcessor : IBatchProcessor
     {
-        public WorkQueue Queue { get; private set; }
```
Noggog: This change was inspired because the GUI currently maps/displays the current num cores by watching the WorkQueue observable, which it reaches via this property.  By not being set initially (and not implemented INotifyPropertyChanged, as it's not a GUI object), the late set inside `ConfigureProcessor` was not updating the GUI to be aware that the previously null property was now set.

By being set immediately, and instead just calling `SetActiveThreadsObservable`, it avoids this problem naturally

Wabbajack.Lib/ABatchProcessor.cs
```diff
@@ -47,13 +48,18 @@ public abstract class ABatchProcessor : IBatchProcessor
 
         private readonly CompositeDisposable _subs = new CompositeDisposable();
 
-        protected void ConfigureProcessor(int steps, int threads = 0)
+        // WorkQueue settings
+        public BehaviorSubject<bool> ManualCoreLimit = new BehaviorSubject<bool>(true);
+        public BehaviorSubject<byte> MaxCores = new BehaviorSubject<byte>(byte.MaxValue);
+        public BehaviorSubject<double> TargetUsagePercent = new BehaviorSubject<double>(1.0d);
```
Noggog: Not too happy on the presence of these settings here, but not sure where they should go exactly right now.  Maybe moving them to WorkQueue itself could be a better spot for them.

Wabbajack.Test/WorkQueueTests.cs
```diff
@@ -0,0 +1,273 @@
...
+                bool[] workStartedArray = new bool[Large];
+                async Task Job(int num, bool[] b)
+                {
+                    // Mark work started as soon as job started
+                    lock (lockObj)
+                    {
+                        b[num] = true;
+                    }
+                    // Do lots of hard work for 1 second
+                    Thread.Sleep(5000);
```
Noggog: These two tests aren't really "tests", but rather testable concrete examples of pitfalls to watch out for when using WorkQueue.  Might find them interesting to look over

Noggog:
Alright, I think this PR is ready for merging on my end.  I would still recommend looking the WorkQueue side over

Wabbajack.Common/WorkQueue.cs
```diff
@@ -27,64 +30,139 @@ public class WorkQueue : IDisposable
         private readonly Subject<CPUStatus> _Status = new Subject<CPUStatus>();
         public IObservable<CPUStatus> Status => _Status;
 
-        public List<Thread> Threads { get; private set; }
+        private int _nextCpuID = 1; // Start at 1, as 0 is "Unassigned"
```
Noggog: Not that it matters too much, but figured I'd point out that the CPU ID is now not `0-(num cores)`, but 0 and increasing to any value (which may be higher than the current number of active cores).  The logic was a bit simpler this way, and I don't think anyone was leveraging CPU ID in a way that required it to be within `0-(num cores)`.

halgari:
Looks good to me, Rx really makes the Work queue management elegant. And I like that we're getting away from dedicated threads in work queues.

PR #397, opened 2020-01-18
Allow Wabbajack to work without Nexus Premium.

Unnoen:
If the Nexus user does not have premium, the archive will fall back to the ManualDownloader.
Remove some unnecessary logging from ManualDownloader.
Log how many archives remain.

### DIFF
--- a/Wabbajack.Lib/AInstaller.cs
+++ b/Wabbajack.Lib/AInstaller.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Alphaleonis.Win32.Filesystem;
 using Wabbajack.Common;
 using Wabbajack.Lib.Downloaders;
+using Wabbajack.Lib.NexusApi;
 using Wabbajack.VirtualFileSystem;
 using Directory = Alphaleonis.Win32.Filesystem.Directory;
 using File = Alphaleonis.Win32.Filesystem.File;
@@ -233,6 +234,22 @@ namespace Wabbajack.Lib
 
             Info("Getting Nexus API Key, if a browser appears, please accept");
 
+            var client = await NexusApiClient.Get();
+            bool premium = await client.IsPremium();
+
+            //If we aren't premium, fall back to manual downloading method by converting states.
+            if (!premium)
+            {
+                Utils.Log($"Nexus user {await client.Username()} is not a premium account. Falling back to manual downloads.");
+                foreach (var a in missing.Where(a => a.State.GetType() == typeof(NexusDownloader.State)))
+                {
+                    a.State = new ManualDownloader.State
+                    {
+                        Url = (a.State as NexusDownloader.State).NexusURL
+                    };
+                }
+            }
+
             var dispatchers = missing.Select(m => m.State.GetDownloader()).Distinct();
 
             await Task.WhenAll(dispatchers.Select(d => d.Prepare()));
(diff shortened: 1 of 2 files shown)


Wabbajack.Lib/AInstaller.cs
```diff
@@ -233,6 +234,22 @@ public async Task DownloadArchives()
...
             Info("Getting Nexus API Key, if a browser appears, please accept");
 
+            var client = await NexusApiClient.Get();
+            bool premium = await client.IsPremium();
+
+            //If we aren't premium, fall back to manual downloading method by converting states.
+            if (!premium)
+            {
+                Utils.Log($"Nexus user {await client.Username()} is not a premium account. Falling back to manual downloads.");
+                foreach (var a in missing.Where(a => a.State.GetType() == typeof(NexusDownloader.State)))
```
erri120: Use the `Do` Extension from `Wabbajack.Common.Utils`:
```cSharp
missing.Where(a => a.State.GetType() == typeof(NexusDownloader.State)).Do(f => ....)
```

Noggog: Haha, I've thought a few times of commenting that we should swap `.Do`s for foreach loops. 8P
I think it's pretty heavily personal preference.  I do think a forloop is technically faster as you dont have any lambda variable captures, but that's pretty minor reason to care for most situations.  I'd say it's not much of a problem either way

Noggog:
Hey Unnoen!  Thanks for the PR.

I do think manual download fallback is a feature that would be useful for WJ.  Users without premium would love it, of course, but it would also be nice in times of low/crappy connectivity (Nexus or some other site being slow and unreliable) to give the users the option of trying the downloads themselves and putting their results in easily.

One critique for this PR specifically is probably a side effect of a trap that we laid for you.  **The code your submitting puts too much Nexus-specific logic inside `AInstaller`, which is supposed to be a generic class that should be unaware of what sites/places it's even attempting to interact with.**  The trap we laid for you that I'm mentioning is the existing line 235 that prints "Getting Nexus API Key...".   That makes it sound like your code would be pretty appropriate in this area, but I'd argue it is vestigial and should be removed, rephrased, or moved.  My vote would be to change its phrasing to just "Preparing all logins and API keys.  If browsers appear, please accept".  After cleaning that up, I'd recommend that the code your proposing should probably live in a Nexus-specific subclass like `NexusDownloader`, most likely in its `Prepare` function?

I think that's somewhat of a distraction, though, as **I think every download source should be able to cleanly fall back to manual download techniques if they happen to fail or can't authenticate.**  Whether that's LL, Vector, ModDB, or anything.  So maybe we can put some thought into how to solve this for -all- download sources, which might yield this nexus-specific handling code obsolete?  I'd have to get into gritty specifics of how we'd solve it to speak further, but i think it's something to talk about as a group a little bit before submitting/committing nexus-specific solutions that only work for Nexus alone.

Also, you mentioned on discord: "- The file checking/moving code doesn't attempt retries, instead WJ fails with an error".  That's something I had been daydreaming about **solving via a UserInteraction implementation for failed/manual downloads**.  For reference, UserInteraction systems are the same systems that power the dialog of whether you want to overwrite an existing installation that pops up.   The idea would be that the engine (`NexusDownloader`, say) would find out it couldn't download, or wasn't premium, etc.  It would "fire" a `CouldNotDownload` UserInteraction onto the system, and then wait for its response.  The GUI would sense this had been fired, and show some GUI elements to allow the user to go to the webpage easily, and drag/drop or whatever the results.  When the file had been located, the UserInteraction would be marked successful, and the engine would be able to resume with the results without failing.   It would only actually fail if the user clicked "cancel" or some other button like that.  If this sounds interesting to implement, I can definitely help walk you through how all those systems work and what would be needed.  Just let me know.

TLDR:
 - Your code should be moved to NexusDownloader
 - We might instead, though, want to find a generic solution that solves this problem for all downloaders
 - You might consider the stretch goal of implementing a CannotDownload UserInteraction that hooks into the frontend to get the manual downloads.

Noggog:
Hey @Unnoen, 
We talked a bit on the discord, and it sounded like you wanted to back off this PR?  Let me know, otherwise we'll probably close it soon

PR #410, opened 2020-01-21
Random refactors

Noggog:
Was doing some investigations into the no-match errors last night.  As expected of a refactor addict, I refactored some stuff while looking around.
- Removed the log message in `DropAll` compilation step.  Compilers are printing any no matches right after this is complete, so this line was resulting in duplicate prints.
- No match printing limited to 10 in the GUI.  The rest continue printing in the file logs.  (last nights build was no-matching almost everything, so printing everything to the GUI in such a spammy nature froze things for a long time)
- Refactored no match logic handling to ACompiler for reuse in MO2/Vortex subsystems
- Refactored raw string usage of ".meta" to `Consts.MetaFileExtension`
- `ExtensionManager` renamed to `ModListAssociationManager`.  Const strings moved to `Consts`
- Renamed `ExtensionManager.Extension` to a more appropriate `ModListExtension`
- Added log line to print how many metas are being inferred

**No fix was found/addressed.  These are just refactors while looking around**


halgari:
@Noggog I think the file rename is messing up the project file somehow. It's complaining that the ExtensionManager class no longer exists

Noggog:
Ah, I'll take a look when I get home.  I suspect the cache server project was unloaded, which probably didn't get the naming refactors as a result.

Noggog:
Yeah, had the cache server unloaded.  Fixed it up

PR #415, opened 2020-01-21
Fix queue size being assigned to 0 on low end machines

tr4wzified:


Wabbajack.Lib/ABatchProcessor.cs
```diff
@@ -80,8 +80,9 @@ public async Task<int> RecommendQueueSize()
             const ulong GB = (1024 * 1024 * 1024);
             // Most of the heavy lifting is done on the scratch disk, so we'll use the value from that disk
             var memory = Utils.GetMemoryStatus();
-            // Assume roughly 2GB of ram needed to extract each 7zip archive, and then leave 2GB for the OS
-            var based_on_memory = (memory.ullTotalPhys - (2 * GB)) / (2 * GB);
+            // Assume roughly 2GB of ram needed to extract each 7zip archive, and then leave 2GB for the OS. If calculation is lower or equal to 1 GB, use 1GB
+            var memory_guess = (memory.ullTotalPhys - (2 * GB)) / (2 * GB);
+            var based_on_memory = (memory_guess <= (1 * GB)) ? (1 * GB) : memory_guess;
```
Noggog: I think the `true` branch of this will result in 1024 threads being returned.  You don't want the `* GB` in the result, I don't think.

Wabbajack.Lib/ABatchProcessor.cs
```diff
@@ -80,8 +80,9 @@ public async Task<int> RecommendQueueSize()
...
             var scratch_size = await RecommendQueueSize(Directory.GetCurrentDirectory());
             var result = Math.Min((int)based_on_memory, (int)scratch_size);
```
Noggog: Also, I think the fix would be better suited after this line.  Just a more simplistic global `Math.Max(result, 1)` call that will make sure any input that results < 1 will be bumped up?  Rather than just fixing the memory result.

tr4wzified:
Fixing the based_on_memory now instead with a Math.Min(). Been a long day x)

Noggog:
Lol, all good.  Just bored at work... coding by proxy... Zzz

tr4wzified:
Good luck over there

Issue #416, opened 2020-01-22
Add support for tesalliance downloads


jdsmith2816:
Now that support for authenticated downloads has been added for LL can we extend that to work for tesalliance?

erri120:
tesalliance is ips4, easy to implement

erri120:
Closed with #420

PR #420, opened 2020-01-22
Added Tesalliance Support

erri120:
#416 
Another IPS4 website... why are all modding sites IPS4?? I mean, good for us!


erri120:
Should've done both sites at once... anyway rebase fixes are in

PR #428, opened 2020-01-26
.NET Core Swapover

Noggog:
- Swapped the WPF app to .NET Core 3.1, and all libraries to .NET Standard 2.1
- Refactored WPF app's styling to match new pre-release MahApps Nuget systems
- Process.Start had to be removed/refactored to a new pattern.  I guess since .NET Core is more detached from Windows, the proper way to start processes changed slightly.
- ACompiler.ShowReport() logic removed and migrated to GUI project
- Added the async/await typical warning tweaks to the global .editorconfig
- Removed vestigial "Debug (no command args)" configuration


Noggog:
Some potential remaining issues.   Some Nugets are now showing warnings that they're built with .NET Framework:
- BunnyCDN
- AlphaFS
- Extended.WPF.Toolkit
- Syroot
- WindowsAPICodePack
- MSTest.TestAdapter

Doesn't necessarily mean failure.  For example, Extended.WPF.Toolkits DoubleSpinner we use in the settings still seems to be working, as well as WindowsAPICodePack's file dialogs, etc.

But, I think the unit test nuget is failing to populate the Test Explorer

erri120:
I think some output paths (mostly in the CI) need to be changed due to the output path being `{projectName}/bin/{platform}/{config}/netcoreapp3.1/`

FlyingWraptor:
As an active BunnyCDN developer I can assure you the storage package is targetting netstandard2.0.
Though you shouldn't worry as you're not using any of it's code as of right now.

Issue #427, opened 2020-01-25
Reactive UI: Bindings and Routing


erri120:
- Data Binding with Rx: https://reactiveui.net/docs/handbook/data-binding/
- Routing with Rx:  https://reactiveui.net/docs/handbook/routing/

_the following was first posted on Discord_
***
so from my testing I'm already liking Rx Routing more than what we have atm:
The thing that throws me off whenever I see WJs WPF XAMLs is the fucking bindings, I know that with DataContext we can get _some_ info in the Editor but using
```csharp
this.WhenActivated(disposable =>
{
  this.Bind(ViewModel, x => x.vmProperty, x => x.viewProperty).DisposeWith(disposable);
});
```
is just soooo good.

Than the routing:
I still have no real idea how `ContentPresenter` works and find it very funky to work with, with RxRouting we make your `MainWindowVM` implement `IScreen` to get `RoutingState`. 
```csharp
Router.Navigate.Execute(new SomethingVM());
```
can get us anywhere.
Ofc we had to swap out the entire `ContentPresenter` block with 
```xml
<rxui:RoutedViewHost
  x:Name="RoutedViewHost"/>
```
and bind that to the `RoutingState Router` in `MainWindowVM` using the RxBindings shown above.
`ViewModels` would also need to implement `IRoutableViewModel` for `UrlPathSegment` and `HostScreen`.
***

How to implement in Wabbajack:

We need to start by changing all UserControls to `reactiveUi:ReactiveUserControl` and the `MainWindow` from `mahapps` to `reactiveUI:Window` (still need to check how that is done @Noggog maybe you got an idea).

`reactiveUi:ReactiveUserControl` references the VM using `x:TypeArguments`. In the `View.xaml.cs` we can bind to the VM using:

```csharp
this.WhenActivated(disposable =>
{
  this.Bind(ViewModel, x => x.vmProperty, x => x.viewProperty).DisposeWith(disposable);
});
```

ViewModel comes from `ReactiveUserControl`/`ReactiveWindow`.

Since routing is done using `Router.Navigate.Execute(new SomethingVM()):` we would need to change the `MainWindowVM.NavigateTo` functions and remove the `ActivePane` and all `Lazy<..VM>` fields.

Our current UI code base is all over the place and making this change will likely require some massive changes but it's well worth it.

Noggog:
We're doing the bindings via `ReactiveUserControl` already.  Maybe there's a few views not converted over, but for the most part it's swapped.
[...]

PR #437, opened 2020-01-27
Reactive Binding + Routing everywhere

erri120:
#427 

- created ReactiveMetro to combine MetroWindow and ReactiveWindow

### DIFF
--- a/Wabbajack/UI/ReactiveMetro.cs
+++ b/Wabbajack/UI/ReactiveMetro.cs
@@ -4,8 +4,23 @@ using ReactiveUI;
 
 namespace Wabbajack.UI
 {
+    /// <summary>
+    /// A <see cref="MetroWindow"/> that is reactive.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This class is a <see cref="MetroWindow"/> that is also reactive. That is, it implements <see cref="IViewFor{TViewModel}"/>.
+    /// You can extend this class to get an implementation of <see cref="IViewFor{TViewModel}"/> rather than writing one yourself.
+    /// </para>
+    /// </remarks>
+    /// <typeparam name="TViewModel">
+    /// The type of the view model backing the view.
+    /// </typeparam>
     public class ReactiveMetro<TViewModel> : MetroWindow, IViewFor<TViewModel> where TViewModel : class
     {
+        /// <summary>
+        /// The view model dependency property.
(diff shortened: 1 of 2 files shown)


Wabbajack/UI/ReactiveMetro.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Windows;
+using MahApps.Metro.Controls;
+using ReactiveUI;
+
+namespace Wabbajack.UI
```
Noggog: Maybe this is for another discussion, but I'm usually not of the opinion that classes' namespaces should match their contained folders.  I feel like this leaders to either excessive namespace definitions... or an unorganized codebase as you try to simplify the namespaces by putting -everything- in the same folder.

Especially in an end-use project like this, I tend to just use one namespace despite using organization folders.  So I'd recommend swapping the namespace to `Wabbajack`?

Wabbajack/UI/ReactiveMetro.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public class ReactiveMetro<TViewModel> : MetroWindow, IViewFor<TViewModel> where TViewModel : class
+    {
+        public static readonly DependencyProperty ViewModelProperty =
+            DependencyProperty.Register(
+                "ViewModel",
```
Noggog: Might've just been copy/pasted from the RxUI website like this, but I'd recommend `nameof(ViewModel)`

Wabbajack/UI/ReactiveMetro.cs
```diff
@@ -4,23 +4,43 @@
 
 namespace Wabbajack.UI
 {
+    /// <summary>
+    /// A <see cref="MetroWindow"/> that is reactive.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This class is a <see cref="MetroWindow"/> that is also reactive. That is, it implements <see cref="IViewFor{TViewModel}"/>.
+    /// You can extend this class to get an implementation of <see cref="IViewFor{TViewModel}"/> rather than writing one yourself.
```
Noggog: Maybe take this one comment line out?  I don't expect anyone to inherit from MetroWindow too often.  Any other view that wants `IViewFor` probably won't be a window and should just inherit `ReactiveUserControl`, or one that only implementes IViewFor, without getting the MetroWindow logic?

Issue #440, opened 2020-01-28
Out of Memory error


halgari:
**Describe the bug**
A user got this error, we should look into why.

```
3393.82 - Extracting Armoury of Tamriel v2.0-29346-2-0-6-1579784416.zip - 100%
3425.3 - Compiler error - System.OutOfMemoryException: Insufficient memory to continue the execution of the program.
   at Wabbajack.Common.Utils.ErrorThrow(Exception ex, String extraMessage) in d:\work\1\s\Wabbajack.Common\Utils.cs:line 100
   at Wabbajack.Common.FileExtractor.<ExtractAll>d__2.MoveNext() in d:\work\1\s\Wabbajack.Common\FileExtractor.cs:line 56
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Wabbajack.VirtualFileSystem.VirtualFile.<Analyze>d__54.MoveNext() in d:\work\1\s\Wabbajack.VirtualFileSystem\VirtualFile.cs:line 157
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Wabbajack.Common.Utils.<>c__DisplayClass60_1`2.<<PMap>b__1>d.MoveNext() in d:\work\1\s\Wabbajack.Common\Utils.cs:line 667
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Wabbajack.Common.Utils.<PMap>d__60`2.MoveNext() in d:\work\1\s\Wabbajack.Common\Utils.cs:line 690
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #455, opened 2020-01-30
Tell 7zip not to use so many threads

halgari:
During decompression 7zip is going wild with threads causing some archives to take as much as 4GB during extraction. Switching this feature off saw my memory usage during extraction of these files to go down to 100MB or so


halgari:
fixes #440

PR #458, opened 2020-01-31
VFS QoL Changes by an angry boi

erri120:
This is me not wanting to have WJ fail after more than 7h of indexing without writing to VFS once in a while.

Changes:

1) formatting and some syntax changes
2) WJ will not index a non-existing file
3) WriteToFile periodically

2nd Point:
Even tho we already filter non existing files out:
`var filtered = Index.AllFiles.Where(file => File.Exists(file.Name)).ToList();`
It is possible that a file in the 
`var filesToIndex = Directory.EnumerateFiles(root, "*", DirectoryEnumerationOptions.Recursive).Distinct().ToList();` List get's deleted before being enumerated upon. Adding a simple check, returning null and than filtering out the null values prevents this.

3rd Point:
Every 10000th file, we write the VFS to file. This worked during my testing of ~30k files but I see a problem with multiple threads trying to change the same variable and than checking it afterwards:

```csharp
var count = 0;
var allFiles = await filesToIndex.PMap(Queue, async f => {
//...
count++;
//...
if (count != 0 && count % AutoSave == 1)
  await WriteToFile
//...
```

I only have a max of 8 threads but someone with 16+ will maybe run into an issue where multiple threads increment `count` and `count % AutoSave == 1` is always false because we count "too fast". Maybe instead of simply calling `count++` we call `var localCount = count++` and use that local variable to compare against.

### DIFF
--- a/Wabbajack.VirtualFileSystem/Context.cs
+++ b/Wabbajack.VirtualFileSystem/Context.cs
@@ -160,34 +160,12 @@ namespace Wabbajack.VirtualFileSystem
             });
         }
 
-        /*public async Task WriteToFile()
+        public void WriteToFile()
         {
-            using (var fs = File.Open(VFSFile, FileMode.Create))
-            using (var bw = new BinaryWriter(fs, Encoding.UTF8, true))
-            {
-                fs.SetLength(0);
-
-                bw.Write(Encoding.ASCII.GetBytes(Magic));
-                bw.Write(FileVersion);
-                bw.Write((ulong) Index.AllFiles.Count);
-
-                (await Index.AllFiles
-                    .PMap(Queue, f =>
-                    {
-                        var ms = new MemoryStream();
-                        f.Write(ms);
-                        return ms;
-                    }))
-                    .Do(ms =>
-                    {
-                        var size = ms.Position;
-                        ms.Position = 0;
-                        bw.Write((ulong) size);
-                        ms.CopyTo(fs);
-                    });
-                Utils.Log($"Wrote {fs.Position.ToFileSizeString()} file as vfs cache file {VFSFile}");
-            }
-        }*/
+            CompositeDisposable.Dispose();
+            _vfsBinaryWriter.Close();
+            _vfsFileStream.Close();
+        }
 
         public async Task IntegrateFromFile(string filename)
         {
--- a/Wabbajack.VirtualFileSystem/Context.cs
+++ b/Wabbajack.VirtualFileSystem/Context.cs
@@ -22,12 +22,13 @@ namespace Wabbajack.VirtualFileSystem
         static Context()
         {
             Utils.Log("Cleaning VFS, this may take a bit of time");
-            Utils.DeleteDirectory(_stagingFolder);
+            Utils.DeleteDirectory(StagingFolder);
         }
         public const ulong FileVersion = 0x02;
         public const string Magic = "WABBAJACK VFS FILE";
+        public const int AutoSave = 500;
 
-        private static readonly string _stagingFolder = "vfs_staging";
+        private const string StagingFolder = "vfs_staging";
         public IndexRoot Index { get; private set; } = IndexRoot.Empty;
 
         /// <summary>
@@ -51,10 +52,10 @@ namespace Wabbajack.VirtualFileSystem
 
         public TemporaryDirectory GetTemporaryFolder()
         {
-            return new TemporaryDirectory(Path.Combine(_stagingFolder, Guid.NewGuid().ToString()));
+            return new TemporaryDirectory(Path.Combine(StagingFolder, Guid.NewGuid().ToString()));
         }
 
-        public async Task<IndexRoot> AddRoot(string root)
+        public async Task<IndexRoot> AddRoot(string root, string filename)
         {
             if (!Path.IsPathRooted(root))
                 throw new InvalidDataException($"Path is not absolute: {root}");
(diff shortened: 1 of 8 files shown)


Noggog:
Definitely looking interesting and promising.  We'll definitely want to add some thread safety, though.  

Like you mentioned, depending on coincidence of work being done, a count++ that should do a save might skip if another thread increments quickly before it reaches the save if statement.

Another scarier issue, though, is what happens when/while one thread enters that if statement and starts writing VFS content... while all the others continue and are adding/modifying it.  A good amount of danger present in this possibility.

------

In very simplistic setups where only the count variable needed to be made safe, you can sometimes get away with `Interlocked.Increment(ref count);`, which is a threadsafe `count++`

However, the if statement's presence means we need more security then just that.  We need thread locking around the increment, and the if statement.  So maybe:

```
// Pseudocode, just to give an idea
AsyncLock lockObj = new AsyncLock();
using (await lockObj.Wait())
{
                  if(filesToIndex.Count >= AutoSave)
                        count++;

                    //TODO: possible that this never happens because another thread is incrementing count
                    if (count != 0 && count % AutoSave == 1 && !string.IsNullOrWhiteSpace(filename))
                        await WriteToFile(filename);
}
```

This will make at least the if statement logic reliable (a count increment that should autosave will always autosave now)

------

However, there's still the possibility that a thread gets the lock, increments the count, enters the save statement... all while another thread is executing `return await VirtualFile.Analyze(this, null, f, f, true);`.  That Analyze might modify the VFS system while `WriteToFile` is being executed, which could trigger any number of errors depending on the gritty details of what's happening.  This is the real problem to solve before the PR is ready to go, imo.

Noggog:
My initial suggestion for this would be that we create an `AsyncReaderWriterLock` class, similar to `AsyncLock`, but implementing this pattern:  https://docs.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim?view=netframework-4.8

This pattern allows for "two levels" of locking.  "Read" level locks can exist alongside each other.. but may never exist alongside a "Write" level lock.

The concept would be we'd wrap the Analyze call in a "read" level lock, while wrapping the count/write block in a "Write" level lock.   This way, `WriteToFile` will never be called while `Analyze` is being executed.   But.. if no `WriteToFile` is being run, then many `Analyze` calls can all be running at the same time for parallelization

halgari:
So there's a way around this, if y'all want to do it *right*. The format of the VFS cache is in essence an append-only log: 

```
MAGIC_STRING
FORMAT_VERSION
<number of files>
<file1>
<file2>
...

<fileX>
```

Duplicated files are ignored during ingestion so they have no impact aside from duplicating work. And as stated before, no single thread has access to the entire VFS cache during indexing. So what we *can* do is setup a IObservable<VirtualFile> that does the following:

1) On.next Virtual File
2) Execute on single thread
3) Append Virtual File record to end of vfs cache
4) Update the "file count" field
5) Fsync (flush)

If we ever crash mid-write, we'll lose a maximum of 1 file because the file_count wasn't updated after a file was appended. Partial writes to the end of the cache will be ignored because the file_count hasn't been updated yet.

Noggog:
Would the subject/saving logic be part of the code that this PMap is a part of?  Or would this system you're proposing be put inside the VirtualFile class somewhere as a common feature of the VFS system?

halgari:
As part of the context, whenever a top level VirtualFile indexes, it puts itself on the Context's subject (VirtualFiles reference the context). The context then listens to the subject and streams files to disk.

erri120:
```csharp
public void AppendToVFS(VirtualFile file)
{
    var ms = new MemoryStream();
    file.Write(ms);

    var size = ms.Position;
    ms.Position = 0;
    _vfsBinaryWriter.Write((ulong) size);
    ms.CopyTo(_vfsFileStream);
}
```
with

```csharp
private readonly FileStream _vfsFileStream;
private readonly BinaryWriter _vfsBinaryWriter;
public IObservable<VirtualFile> IndexingObservable;
``` 

```csharp
IndexingObservable.Subscribe(AppendToVFS, () =>
{
   //TODO: onComplete
});
```

and ofc being disposed on close

Noggog:
Hows this all working out?  Seems to be getting closer to the Rx solution just glancing at it?
I'll be adding a few comments here and there.. but I'm curious how the behavior is when actually running it.  Is it failing spectacularly? Or mostly working?

Wabbajack.VirtualFileSystem/Context.cs
```diff
@@ -39,85 +39,99 @@ static Context()
...
+            VFSFile = vfsCacheFile;
+
+            CompositeDisposable = new CompositeDisposable();
 
+            _vfsFileStream = File.Open(VFSFile, FileMode.Create);
+            _vfsBinaryWriter = new BinaryWriter(_vfsFileStream, Encoding.UTF8, true);
+
+            IndexingObservable.Subscribe(AppendToVFS, () =>
+            {
+                //TODO: onComplete
```
Noggog: Don't necessarily need an on complete lambda here.  This won't actually be called, actually, unless the `.Complete()` call is made on the `Subject`.

Even, then, though, I don't think there's a specific action that needs to be taken once all the items are processed and no more are coming down the pipe?

Wabbajack.VirtualFileSystem/Context.cs
```diff
@@ -146,33 +160,11 @@ class Box<T>
...
-                    .Do(ms =>
-                    {
-                        var size = ms.Position;
-                        ms.Position = 0;
-                        bw.Write((ulong) size);
-                        ms.CopyTo(fs);
-                    });
-                Utils.Log($"Wrote {fs.Position.ToFileSizeString()} file as vfs cache file {filename}");
-            }
+            CompositeDisposable.Dispose();
```
Noggog: This might not be doing what you think it is.  This disposal will unsubscribe the subscription on line 65, so that it no longer exists or does any work.  It will stop the subscription work immediately (maybe after finishing its current item?), -without- clearing the remaining items on the "queue".

Disposal like this is for when we're all done, and want all the subscriptions cleaned up and destroyed... which I imagine is not what you are after here?

If you're after a "hey, don't' expect any more items.. and finish all that you have" logic.. you'll probably want something like:
```
// Tell the subject/observable that there are no more items coming down the line.
// If/when the last item is processed, the observable is now "complete" and any onComplete callbacks
// will at that point fire.
_nextIndexedFile.Complete();

// We've marked the subject as complete, but there's still a lot of items in the pipe, potentially.
// This can be used to "wait" for all the items to be processed, even if they're all being done on different threads or whatever.
// This of course requires the WriteToFile() to be made an async function again
await _nextIndexedFile;

// Now that we've marked the observable complete, and waited for all the items to be finished... now we can shut down and return
CompositeDisposable.Dispose(); // Destroy all subscription wiring
_vfsBinaryWriter.Close();
_vfsFileStream.Close();

File: Wabbajack.VirtualFileSystem/Context.cs (hunk 1 above)
Comment (Noggog) [reply Wabbajack.VirtualFileSystem/Context.cs]: Also, I would consider adding an
```.ObserveOn(RxApp.TaskpoolScheduler)```
So that the work of saving to disk in `AppendToVFS` is done on some random thread, not on the same thread that fired the subject

File: Wabbajack.VirtualFileSystem/Context.cs [hunk 3]
```diff
@@ -336,7 +328,7 @@ public class KnownFile
 
     public class DisposableList<T> : List<T>, IDisposable
```
Noggog: Random side tangent, but might be nice to start moving these classes which are pretty reusable and might be helpful for other code.. into a `Utility` folder?   Maybe `Wabbajack.Common/Util`?

Issue #518, opened 2020-02-14
Have different vfs_compile_cache files for different games


erri120:
Instead of one `vfs_compile_cache.bin`, have `vfs_compile_cache-{Name}.bin` with `Name` being eg the nexux mods game name.

PR #527, opened 2020-02-14
Fix compilation network issues

halgari:
We were (and still are in some cases) abusing HttpClient. This code switches to a singleton HttpClient (which pools connections) and is backed by the new SocketsHttpHandler.


halgari:
also fixes #518

PR #528, opened 2020-02-15
Associate mod list files with the executable instead of a DLL

LostDragonist:
Reorganizing the program to be lots of small files instead of one monolithic
executable changed the output of System.Reflection.Assembly.GetExecutingAssembly
to Wabbajack.dll instead of Wabbajack.exe.

This fix assumes the DLL containing the calling code is in the same folder
as Wabbajack.exe.  There's room for improvement if anyone can figure out how
to positively identify Wabbajack.exe.

### DIFF
--- a/Wabbajack.Common/ModListAssociationManager1.cs
+++ b/Wabbajack.Common/ModListAssociationManager1.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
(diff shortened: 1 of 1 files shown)


LostDragonist:
I'm calling shenanigans on that test failure.  The download test for vectorplexus.com failed.  The code touched nothing related to the downloader and the test works fine locally (... really...).

halgari:
Yeah, their entire site is down right now

PR #529, opened 2020-02-15
Print VFS Cache location

erri120:


erri120:
lol nvm had some old commit in here

PR #574, opened 2020-02-24
Add non-Nexus mods to the Slideshow

### LINKED ISSUE #573: Add non-Nexus mods to the Slideshow

erri120:
resolves #573


Wabbajack/View Models/ModVM.cs
```diff
@@ -13,37 +13,39 @@ namespace Wabbajack
...
+            State = state;
+
+            URL = state.URL;
+            ImageURL = state.ImageURL;
+            IsNSFW = state.IsNSFW;
+            Version = state.Version;
+
+            var isNexus = state.GetType() == typeof(NexusDownloader.State);
+            
+            Name = isNexus ? NexusApiUtils.FixupSummary(state.Name) : state.Name;
```
Noggog: I think ModVM should be agnostic of whether it's nexus or not.  the IMetaState interface should be doing any fixup needed internally and exposing the correct name, i think.

Noggog:
I think https://github.com/wabbajack-tools/wabbajack/blob/836e23537391ed7f3218bd1d21faa5ddc6f4d917/Wabbajack/View%20Models/SlideShow.cs#L90 will want to be changed to be the new interface?

Wabbajack/View Models/ModVM.cs
Noggog: Just to clarify, I think while the `State` class implements `IMetaState` and has a `Name` property... NexusState can override it and provide the fixup itself internally.

Either by actually doing it to the `Name` property straight up, or alternatively, can do the fixup when `Name` is accessed through the `IMetaState.Name` member only, like this:
```
// Property still exists on the base class like normal
string Name { get; set; }

// Intercept and override the behavior for IMetaState.Name calls only
string IMetaState.Name
{
   get => NexusApiUtils.FixupSummary(this.Name);
   set => this.Name = value;
}
```

Wabbajack.Lib/Downloaders/AbstractDownloadState.cs
```diff
@@ -7,6 +7,17 @@
 
 namespace Wabbajack.Lib.Downloaders
 {
+    public interface IMetaState
```
Noggog: Might consider just making this readonly?  Eg.. removing the `set` halves.
I imagine most users of this interface just want to get things.  Might make things simpler, like with the NexusState overriding the `Name` property.. it doesn't also have to define the set side

Wabbajack/View Models/ModListVM.cs
```diff
@@ -38,6 +38,7 @@ public ModListVM(string modListPath)
             catch (Exception ex)
             {
                 Error = ex;
+                Utils.Log($"Error while loading the modlist:\n{ex}");
```
Noggog:
```suggestion
                Utils.Error(ex, "Error while loading the modlist");
```
I think this would be preferable, as it gives us a concrete way of identifying which messages are "errors", vs all of them just being log lines that might be errors.. but might just be info.

If we want to eventually update the WJ log display, for ex, it would be useful to know which lines are errors, so they can be displayed red, or whatever... that sorta idea

erri120: right, force of habit

Wabbajack.Lib/Downloaders/AbstractDownloadState.cs
```diff
@@ -7,21 +7,23 @@
...
-        string Description { get; }
+        public abstract string URL { get; }
+        public abstract string Name { get; set; }
+        public abstract string Author { get; set; }
+        public abstract string Version { get; set; }
+        public abstract string ImageURL { get; set; }
+        public abstract bool IsNSFW { get; set; }
+        public abstract string Description { get; set; }
+
+        public abstract Task<bool> LoadMetaData();
```
Noggog: I think I agree with @halgari a bit.  If it's just listing members/functions that should exist, then using an interface is much cleaner, as it better supports "multi-inheritence" of many interfaces.  Having it as a base class only is useful if it also implements something /w some logic.   As it stands, it's just declaring things that need to exist, so an interface is sufficient.

Wabbajack/View Models/SlideShow.cs
```diff
@@ -82,23 +82,23 @@ public SlideShow(InstallerVM appState)
                 {
                     if (modList?.SourceModList?.Archives == null)
                     {
-                        return Observable.Empty<NexusDownloader.State>()
-                            .ToObservableChangeSet(x => x.ModID);
+                        return Observable.Empty<AbstractMetaState>()
+                            .ToObservableChangeSet(x => x.Name);
```
Noggog: This will result in any two mods with the same name colliding (and replacing one another, I think?).    Probably not an issue in reality, I imagine, but it's a theoretical issue.   Name is probably a fine enough key for real life.

Two ways to solve the theoretical problem if we want to:
1)  Find a way to uniquely identify all mods (before when it was nexus, the ID was enough.  With cross-site mods, though, no longer)
2)  I think we could actually just remove the `Cache`/`Dictionary`/`Key` concepts and just use list mechanics.  Basically.. we assume the modlist doesn't have duplicate entries.  Swap to `.ToObservableChangeSet()` with no key specified.
Would have to try and test.

Wabbajack/View Models/SlideShow.cs
```diff
@@ -82,23 +82,23 @@ public SlideShow(InstallerVM appState)
...
-                        .OfType<NexusDownloader.State>()
+                        .OfType<AbstractMetaState>()
                         // Shuffle it
                         .Shuffle(_random)
-                        .AsObservableChangeSet(x => x.ModID);
+                        .AsObservableChangeSet(x => x.Name);
                 })
                 // Switch to the new list after every ModList change
                 .Switch()
                 .Transform(nexus => new ModVM(nexus))
```
Noggog:
```suggestion
                .Transform(mod => new ModVM(mod))
```

Noggog:
PR looks good to me.  Only real qualm I have is the base class usage.

Wabbajack.Lib/Downloaders/AbstractDownloadState.cs
Noggog: Just "bumping" this comment, as it's applicable again.  I think the interface would be better as a get-only interface.  No need to require implementers to provide a `set`.  Users of `IMetaState` will likely only ever be reading info.. not setting it.

The reason this might be important is that requiring a `set` makes it difficult for the classes implementing the interface if they'd prefer the fields to be derivative (eg, driven by some other internal data that is queried/processed to give the result).

Something like this wouldn't work if set was required:
```
public string Name => someDataStructure.Find("name").Value;
```

PR #579, opened 2020-02-25
Fixed ArgumentException in Octodiff Reporter

erri120:
Fixed [this](https://cdn.discordapp.com/attachments/620954238931173378/681810078256332834/Wabbajack.current.log) problem


Noggog:
Made a discord comment about this, but `Utils.Status(operation, Percent.FactoryPutInRange(total, currentPosition));` is what you're after, I think.

The exception happening suggests your currentPosition/total calculations are wrong somehow, but at least they'll be put into a viable 0-1 range with the above call, rather than an exception

We could also do a rename to `Percent.PutInRange` while we're here.  I think the `Factory` is a bit verbose, haha

PR #609, opened 2020-03-05
Use virtual memory during BSA creation

### LINKED ISSUE #606: BSAs keep their contents in memory during creation
**Describe the bug**
During construction, BSAs keep their entire contents in-memory. This is to reduce open file handles, and also because the size of a compressed BSA entry isn't known until the data is compressed. 

**Additional context**
FO4 BA2 files can be around 10GB compressed. I've seen usage as high as 16GB during BA2 creation. Once the BA2 is written this memory is released (showing it's not a memory leak) but it's still way too much memory. And we try to support 16GB systems as much as possible.

**Solution**
I think we can leverage MemoryMappedFiles to cache data outside of the main memory. 64bit systems support roughly 16mil file handles open at a time in a given process, so I think we should be fine to leave uncompressed data on-disk and only read it in during the writing of the actual BSA file.

We could also use temporary files to store the compressed data, but that would result in potentially tens of thousands of small files existing on disk. Memory mapped files allow us to keep all the compressed data in one location.

halgari:
Fixes #606


halgari:
I want to fix the other BSA types before I merge this

PR #611, opened 2020-03-05
More Slideshow Settings

erri120:
Created new panel in the settings view:
05-03-2020 16-50-53 Wabbajack

I (will) use this to debug the recent PR #574 . Since we already have a _Show NSFW_ button in the slideshow directly I maybe remove the `Allow NSFW` button.

What I want to also add is a selector for `IMetaStates` you want to see, eg: only show mods from NexusMods/LL/whatever for better debugging and maybe someone wants to know.


Noggog:
I think this is a PR that will open a can of worms.  There are so many very specific situations of things we need/want to test in WJ...  so many specific one-off items to test for.  If we make settings for everything we want to test, then it'll get massive and messy real quick, I think.

Would it not be easier to just make a test modlist /w only the items you want to see in the slideshow?

PR #614, opened 2020-03-06
Added ProcessHelper.cs and Paths.cs

halgari:


halgari:
I'm going to scrap this and break the PR up a bit, what started as a simple refactor ended up being way too extensive

PR #663, opened 2020-04-01
Pathing Refactors

Noggog:
The main inspiration for digging around was that I noticed that all the paths were `ToLower`ed.  This affected a lot of stuff including:
- Being displayed weirdly and overcorrecting typed input on the GUI to always be lower
- Affecting items populated by paths such as automatically populated ModList names based on Mod Organizer folder names.

The across the board `ToLower`ing I think will/would cause random annoying problems like those in a lot of unexpected places.

I refactored the Pathing systems to not `ToLower` during construction, and keep the string path as given.  Instead, they make use of string-specialized case-ignoring Hash/Equals comparisons to maintain the case-insensitivity requirements.

Other changes:
- AbsolutePath implements IEquatable.  Already was, but just added the interface
- AbsolutePath constructors merged
- Modified `Extension` members of Absolute/RelativePaths to not calculate during construction, but rather on demand when asked for.  I think this the more desired/expected behavior?

### DIFF
--- a/Wabbajack.Common/Paths.cs
+++ b/Wabbajack.Common/Paths.cs
@@ -55,7 +55,6 @@ namespace Wabbajack.Common
         public AbsolutePath(string path, bool skipValidation = false)
         {
             _path = path.Replace("/", "\\").TrimEnd('\\');
-            Extension = Extension.FromPath(path);
             if (!skipValidation)
             {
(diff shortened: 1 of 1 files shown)


halgari:
Thank you so much for looking into this, I agree the variant without lowering is better.

PR #669, opened 2020-04-02
PR: Add MEGA Login

### LINKED ISSUE #664: Add MEGA Login
We currently auto login as anonymous when downloading something from MEGA ([reference](https://github.com/wabbajack-tools/wabbajack/blob/master/Wabbajack.Lib/Downloaders/MEGADownloader.cs#L31)). This is fine for most stuff but the reason mods are being hosted on MEGA is because they are huge. If a modlist only has one or two MEGA links, all is fine but when you have something like [Lotus](https://github.com/wabbajack-tools/opt-out-lists/blob/master/ServerWhitelist.yml#L177) which downloads multiple big mods from MEGA, not only will we get reduced speed but also reach the download limit soon-ish for the user.

If the user has a MEGA account we could add MEGA to the login manager and login using other `MegaClient.Login()` overloads ([ref](https://gpailler.github.io/MegaApiClient/api/CG.Web.MegaApiClient.MegaApiClient.html#CG_Web_MegaApiClient_MegaApiClient_Login_CG_Web_MegaApiClient_MegaApiClient_AuthInfos_)).

erri120:
Resolves #664

### DIFF
--- a/Wabbajack.Common/Utils.cs
+++ b/Wabbajack.Common/Utils.cs
@@ -11,6 +11,7 @@ using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Security;
 using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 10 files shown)


erri120:
Login and the UI are working, only remaining thing is implementing some form of custom `PasswordBox` or `TextBox`. `PasswordBox` has `string Password` and `SecureString SecurePassword` which are both not bindable.

erri120:
I swapped the `TextBox` to `PasswordBox` and not went with RXUI bindings but with your classic XAML bindings. This might not be ideal for a RXUI based application but it's by far the quickest and cleanest solution.
Tested login, logout, entire compilation + installation.

PR #670, opened 2020-04-02
Nullability Flavor Bootcamp

Noggog:
I feel like the internal refactoring pass of 2.0 is a good time to transition to Nullability concepts.  Figured it swap it on and get one of the libraries swapped over as an example/pitch.

This swaps Wabbajack.Common over to Nullability enabled.  Everything should function the way it did before.  Some small refactors were needed to make things conform to the nullability concepts (initializations moved to ctors, for ex).


Wabbajack.Common/DynamicIniData.cs
```diff
@@ -35,7 +36,7 @@ public override bool TryGetMember(GetMemberBinder binder, out object result)
             return true;
         }
         
-        public override bool TryGetIndex(GetIndexBinder binder, object[] indexes, out object result)
+        public override bool TryGetIndex(GetIndexBinder binder, object[] indexes, [MaybeNullWhen(false)] out object result)
```
Noggog, line 39: This is an example of MaybeNullWhen concepts for use in TryGet concepts.
This allows you to assign null to a normally non-nullable out parameter.  The compiler also notices this on the user-end and will warn the user when they use the variable without checking the return bool first.

Wabbajack.Common/Extensions/DictionaryExt.cs
```diff
@@ -6,17 +6,37 @@ namespace Wabbajack
...
+            where K : notnull
         {
             if (dict.TryGetValue(key, out var val)) return val;
             var ret = create();
             dict[key] = ret;
             return ret;
         }
+
+        public static void Add<K, V>(this IDictionary<K, V> dict, IEnumerable<KeyValuePair<K, V>> vals)
+            where K : notnull
```
Noggog, line 25: **Example of `notnull` generic constraints.**
Dictionaries can never have null in their key collection, so this constraint ensures the generic can never be a nullable type.

Wabbajack.Common/Json.cs
```diff
@@ -72,7 +72,7 @@ public static T FromJSON<T>(this Stream data)
             try
             {
                 return JsonConvert.DeserializeObject<T>(s,
-                    new JsonSerializerSettings {TypeNameHandling = TypeNameHandling.Auto, Converters = Converters});
+                    new JsonSerializerSettings {TypeNameHandling = TypeNameHandling.Auto, Converters = Converters})!;
```
Noggog, line 75: **Example of overriding a null warning/error and pushing it through anyway**
I -think- the json API is bad, in that it returns a nullable object, but it would never do so.  It would either return an object, or throw an exception.  But the fact that it returns a nullable object makes the compiler complain when we try to return it in a non-nullable value.  We can override its complaining with `!`

Wabbajack.Common/Paths.cs
```diff
@@ -264,7 +266,7 @@ public AbsolutePath AppendToName(string toAppend)
 
         public AbsolutePath Combine(params RelativePath[] paths)
         {
-            return new AbsolutePath(Path.Combine(paths.Select(s => (string)s).Where(s => s != null).Cons(_path).ToArray()));
+            return new AbsolutePath(Path.Combine(paths.Select(s => (string)s).Cons(_path).ToArray()));
```
Noggog, line 269: **Example of speed upside to adding nullability**
Since the variable contains types that can never be null, we can omit the check for nulls from our logic.

Wabbajack.Common/Paths.cs
```diff
@@ -378,23 +380,24 @@ public async Task CopyDirectoryToAsync(AbsolutePath destination)
 
     public struct RelativePath : IPath, IEquatable<RelativePath>, IComparable<RelativePath>
     {
-        private readonly string _path;
+        private readonly string? _nullable_path;
```
Noggog, line 383: **Structs are dangerous to have non-nullable members**
Structs are allowed to have non-nullable members.  However, there's a danger here that's not caught by the compiler (that i've seen).
The code will still block you from taking actions or assigning things poorly in the code.. but there's a use case that isn't caught:
A `default` struct will have its non nullable members set to null.  A user will then access the member without checking (as its marked not nullable), and then get a null exception.

It's my understanding so far that structs shouldn't have nullable refs, due to them being able to hold null values if `default` is used.

So instead.. we store a nullable path, and have an internal member that checks and returns empty string in that case.
Thus a `default(RelativePath)` will not fail, despite its internal string member containing null.

Nullability concepts are pretty new.. so my understanding of this issue might evolve and/or they may address this later.

Wabbajack.Common/Utils.cs
```diff
@@ -673,7 +666,9 @@ public static T Last<T>(this T[] a)
             return a[a.Length - 1];
         }
 
+        [return: MaybeNull]
```
Noggog, line 669: **Example of `MaybeNull` return attribute to power GetOrDefault Concepts**
Without the attribute, the code will complain when returning `default`.
However, if you swap `V` to be `V?`, then it requires you constrain it to either `class` or `struct`. 
The `MaybeNull` attribute is the way to get around this and set up a proper GetOrDefault setup

Noggog:
Closing to redo onto the rebase

Noggog:
Did a force push with things cherry picked.  Should be good?

Noggog:
Eh, tried to do a force push of cherry picked commits on top of the rebase... but Github seems thoroughly confused about it all still.

PR #701, opened 2020-04-10
Wabbajack.Lib Nullability Enabled

Noggog:
- Migrates Wabbajack.Lib to Nullability concepts
- ABatchProcessor.ConfigureProcessor removed in favor of a pattern where most of the fundamental objects are created in ctor.  This helps keeps the internal members marked non-nullable, as there's never a point in the lifetime that they're null and uninitialized.
- TryGetGameLocation and TryGetByFuzzyName added, which offer nullable return values, vs their non-Try counterparts that throw if not found, but return non-null always.
- Added ctors to some data container classes, rather than just having public properties.  This helps keeps the internal members marked non-nullable, as there's never a point in the lifetime that they're null and uninitialized.
- `DictionaryExt.SetTo`, `ListExt.SetTo` added
- EnumerableExt.NotNull added.  This is very useful for converting an enumerable that has a nullable type to a non nullable type.  Eg:
```
Enumerable<T?> myEnum = ...;
Enumerable<T> filteredAndProperlyTyped = myEnum.NotNull();
```

A simple `.Where(e => e != null)` call doesn't work, as the Enumerable's type would still be `T?`, where all the values coincidentally weren't null, despite their type allowing them to potentially be.

----------

### Some questions remaining:
- JSON serialization.  Their deserialization calls return `T?`, which has been solved by overriding the error `!`.  This is probably okay as is, until they upgrade the library?
- JSON serialization of non-null properties.  For example:
```
public class MyData
{
    string MyString = string.Empty;
}
```
We've got our data class, and we have a string, which should never be null, and we initialize it to not be null.  Compiler is happy, and the code we write won't ever have that string be null without an error popping up.

However... the JSON serialization library might not respect that, and might fill that value to be null if that field doesn't exist in the JSON, rather than filling it to `string.Empty` for us.  This would/could then lead to null ref errors in certain situations.

Maybe not the biggest problem, as any resulting null ref errors would've been triggered before anyway, but something to be aware of.


Noggog:
Large refactor.  Did my best to keep parity of features 1:1 as I worked, but you never know.  Part of the reason I'm pushing hard to do this now, as we'll probably be doing a large amount of testing at this phase, which this can benefit from

Noggog:
Not sure of the state of the unit tests.  Let me know if the failures are expected or not 8)

halgari:
The TESAlliance site goes down (looks like that's currently the case) so ignore that test for now. However we do have about half a dozen errors from "this key already exists in the dictionary", those are normally caused by some check that used to say `result != null then insert into dictionary...`, but now `result` is a value so it's never null so the logic changed. Those errors will need to be fixed.

Noggog:
Should be a bit better.  The dictionary issue was actually just the two-phase HashArchives system (MO2Installer L109-117).  The second call always threw an exception since dictionaries' `Add` call was being used which throws.  Just swapped it to a set

PR #702, opened 2020-04-10
Missing IDownloader.Prepare awaits

Noggog:
Noticed IDownloader.Prepare wasn't being awaited everywhere it was called.  Added most of them.

The only oddball worth talking about is DownloadDispatcher.GetInstance's call of Prepare.   GetInstance is not async, and I'm not sure we want to make it async, as that would affect a large swath of its users?  It also seems/feels like most callers of GetInstance are also calling the downloader's prepare function themselves, and then awaiting that.

So GetInstance's call of Prepare almost feels like it's a good candidate for `FireAndForget`?

Still, it doesn't quite feel right to me because:
1)  If the user of GetInstance doesn't call prepare themselves, then there's potential that the prepare that was called (by GetInstance itself) isn't finished yet by the time they start using it.  This could lead to odd threading errors.
2)  If the user of GetInstance DOES call prepare themselves, then at least **one** prepare was awaited and completed, yes.  But it's also uncertain what the side effects of two prepares on the same object being called are, one of which wasn't awaited and (in theory) could still be trying to run in the background and messing things up.  The burden is then on the downloader to be able to handle multiple concurrent calls to its Prepare.  Maybe they already are good for this?

My suggestion is either:
1)  Make GetInstance async, so that its Prepare can be properly awaited.  (lots of code would then have to be modified to await it)
2)  Don't prepare within GetInstance at all.  Put that burden on the user every time.

Not sure exactly what the right call is.

### DIFF
--- a/Wabbajack.BuildServer/Models/Jobs/UpdateModLists.cs
+++ b/Wabbajack.BuildServer/Models/Jobs/UpdateModLists.cs
@@ -44,7 +44,7 @@ namespace Wabbajack.BuildServer.Models.Jobs
             return JobResult.Success();
         }
         
-         private async Task ValidateList(SqlService sql, ModlistMetadata list, WorkQueue queue, ValidateModlist whitelists)
+        private async Task ValidateList(SqlService sql, ModlistMetadata list, WorkQueue queue, ValidateModlist whitelists)
         {
             var modlistPath = Consts.ModListDownloadFolder.Combine(list.Links.MachineURL + Consts.ModListExtension);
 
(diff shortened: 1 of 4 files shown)


halgari:
Prepare is meant to be called right before we either call `Download` or `Verify`, the idea being that some downloaders may need to start some big subprocess (cef) in order to do their work, so we don't want to tell them to do that until we're sure we need to use the downloader. So I don't think we need a call to `Prepare` in the static method at all. We should however make sure that any downloaders that do work in Prepare do so as a fire-once sort of thing, so once they are prepared they never attempt to prepare again.

Noggog:
Cool, should be good to go, then.

PR #703, opened 2020-04-10
Missing DeleteDirectory awaits

Noggog:
Main thing of note was that Context.Stage needed its cleanup action swapped to `Func<Task>`.  This required addition of AsyncDisposableList.


Noggog:
Fixed up the VFS tests a bit more, utilizing xUnits IAsyncLifetime.. which i think is what we've been after with the back and forth?

Noggog:
Weird, it seems to have failed a test that doesn't exist on the `missing-delete-directory` branch.   Not sure how it could be running a new test that only exists on `master`?

halgari:
Yeah, something's off with the tests, I'll look into it.

PR #719, opened 2020-04-17
Extraction improvements

halgari:
BSA archives are now lazily extracted
7Zip extracted archives now only extract the fewest files required.
Audited the uses of .Wait
Lazily init the VFS cleaning


halgari:
Merging despite broken tests because several IPS4 sites are down

PR #766, opened 2020-04-27
Fixed zEditIntegration not finding the correct settings file and profile folder

erri120:


tr4wzified:
Should probably remove that test var

erri120:
lmao

PR #803, opened 2020-05-07
[Feature] added a filter for games

LordOfRhun:
### Description
this add the feature to filter by game names.

### Specifications
The comboBox is feeded by the _Wabbajack.Common.Game_ enum so any newly added games should be in the filter automatically. the filter works by simply comparing the metadata from the game and the selected game from the filter.

### Preview
https://i.imgur.com/lVSK7DB.gifv

my discord tag is _Khamûl#0128_ for communication

### DIFF
--- a/Wabbajack/Views/ModListGalleryView.xaml
+++ b/Wabbajack/Views/ModListGalleryView.xaml
@@ -10,7 +10,7 @@
     xmlns:rxui="http://reactiveui.net"
     xmlns:system="clr-namespace:System;assembly=mscorlib"
     d:DesignHeight="450"
-    d:DesignWidth="800"
+    d:DesignWidth="900"
     x:TypeArguments="local:ModListGalleryVM"
     mc:Ignorable="d">
     <Grid>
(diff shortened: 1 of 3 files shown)


Noggog:
Looking solid!  Initially I was going to comment on maybe keeping the Enum around longer instead of having a list of strings, but given that we need "All" to be an option on the list, that complicates things.  I think the string solution is a good setup for now with the current featureset.

It's looking good to me

Wabbajack/Views/ModListGalleryView.xaml
```diff
@@ -139,4 +151,4 @@
             <iconPacks:PackIconMaterial Foreground="{Binding Foreground, RelativeSource={RelativeSource AncestorType={x:Type Button}}}" Kind="ArrowLeft" />
         </Button>
     </Grid>
-</rxui:ReactiveUserControl>
+</rxui:ReactiveUserControl>
```
Noggog, line 154: Git gets unhappy without an extra new line at the end.  Not sure why.
Smallest of nothing issues.. just happened to notice it was intentionally removed while scrolling through

Noggog:
Maybe add a `const string` member to store "All".  Just makes it more robust if we want to change the term later

LordOfRhun:
I made some changes from your suggestions, and the games names now comes from the enum description attribute as suggested by @erri120 on discord. It looks much better

PR #815, opened 2020-05-09
Updated the contributing guidelines

LordOfRhun:
### Description

* added a link for .net .4.7.2
* added a step
* removed a dead link

I recently cloned the Repo and thought some more information would be beneficial.

I removed a dead link which could probably be replaced by the a new one but i don't really know your coding Style. If you provide a new link i would be happy to replace it.


LordOfRhun:
the repo the dead link was pointing to has been migrated to https://github.com/dotnet/runtime which also include a coding style https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md but i can't be sure it's the same one as i never saw the original

erri120:
.NET 4.7.2 is not even needed anymore. All of our projects are using .NET Core 3.x and the build server is an ASP.NET project.

Issue #481, opened 2020-02-06
Modlist gallery filter/search


Noggog:
Modlist Gallery needs some display improvements:
- Filter by game (combobox and/or on/off game icon 'checkboxes')
- Filter by name.  
- Sort by name/recently updated (maybe not downloads or anything, as we don't want to get runaway top installed lists?)
- Hide lists under maintenance?

Most of these should be debounced and wired into a filter call on the DynamicData observable chain that eventually binds to the display collection, so that things update live

- Easy clear all filters button
- Would also be good to maybe save these settings.  Most users will want to filter to the same game every time they open WJ

Noggog:
Also, do we perhaps want to do a random shuffle of the modlists?  Just to keep one list from always being the top one displayed?  Might be nice to just help all get a fair shake at being downloaded

LordOfRhun:
I am currently working on this
>     * Would also be good to maybe save these settings.  Most users will want to filter to the same game every time they open WJ

Maybe there could be a check box in the settings disable the shuffle which would be enable by default  ?
> Also, do we perhaps want to do a random shuffle of the modlists? Just to keep one list from always being the top one displayed? Might be nice to just help all get a fair shake at being downloaded

LordOfRhun:
> Also, do we perhaps want to do a random shuffle of the modlists? Just to keep one list from always being the top one displayed? Might be nice to just help all get a fair shake at being downloaded
Hey while coding as saw this implemented.

https://github.com/wabbajack-tools/wabbajack/blob/1bbbda4f350c0af12bb71b695e3ee7f58f3e84e4/Wabbajack/View%20Models/Gallery/ModListGalleryVM.cs#L82-L83
But in my experience it doesn't seems to work,

halgari:
Fixed in master

PR #816, opened 2020-05-10
Added persistent filters on exit and more

LordOfRhun:
Based on [Modlist gallery filter/search #481](https://github.com/wabbajack-tools/wabbajack/issues/481)
### Description

* Added persistent filters on exit
* settings to disable the persistent filters
* **fixed a bug** related to game comboBox and OnlyInstalled ( the comboBox is now disabled if OnlyInstalled is checked )

### Preview
https://i.imgur.com/GG8lY7O.jpg

### DIFF
--- a/Wabbajack/Settings.cs
+++ b/Wabbajack/Settings.cs
@@ -21,6 +21,7 @@ namespace Wabbajack
         public double Height { get; set; }
         public double Width { get; set; }
         public InstallerSettings Installer { get; set; } = new InstallerSettings();
+        public FiltersSettings Filters { get; set; } = new FiltersSettings();
         public CompilerSettings Compiler { get; set; } = new CompilerSettings();
         public PerformanceSettings Performance { get; set; } = new PerformanceSettings();
 
@@ -92,6 +93,18 @@ namespace Wabbajack
         public ModManager LastCompiledModManager { get; set; }
         public AbsolutePath OutputLocation { get; set; }
         public MO2CompilationSettings MO2Compilation { get; } = new MO2CompilationSettings();
+    }
+    
+    [JsonName("FiltersSettings")]
+    [JsonObject(MemberSerialization.OptOut)]
+    public class FiltersSettings : ViewModel
+    {
+        public bool showNSFW { get; set; }
+        public bool OnlyInstalled { get; set; }
+        public string Game { get; set; }
+        public string Search { get; set; }
+        private bool _isPersistent = true;
+        public bool isPersistent { get => _isPersistent; set => RaiseAndSetIfChanged(ref _isPersistent, value); }
     }
 
     [JsonName("PerformanceSettings")]
--- a/Wabbajack/Settings.cs
+++ b/Wabbajack/Settings.cs
@@ -99,12 +99,12 @@ namespace Wabbajack
     [JsonObject(MemberSerialization.OptOut)]
     public class FiltersSettings : ViewModel
     {
-        public bool showNSFW { get; set; }
+        public bool ShowNSFW { get; set; }
         public bool OnlyInstalled { get; set; }
         public string Game { get; set; }
         public string Search { get; set; }
         private bool _isPersistent = true;
-        public bool isPersistent { get => _isPersistent; set => RaiseAndSetIfChanged(ref _isPersistent, value); }
+        public bool IsPersistent { get => _isPersistent; set => RaiseAndSetIfChanged(ref _isPersistent, value); }
     }
 
     [JsonName("PerformanceSettings")]
--- a/Wabbajack/View Models/Gallery/ModListGalleryVM.cs
+++ b/Wabbajack/View Models/Gallery/ModListGalleryVM.cs
@@ -65,10 +65,10 @@ namespace Wabbajack
 
             // load persistent filter settings
             settings = MWVM.Settings.Filters;
-            if (settings.isPersistent)
+            if (settings.IsPersistent)
             {
                 GameType = !string.IsNullOrEmpty(settings.Game) ? settings.Game : ALL_GAME_TYPE;
-                ShowNSFW = settings.showNSFW;
+                ShowNSFW = settings.ShowNSFW;
                 OnlyInstalled = settings.OnlyInstalled;
                 if (OnlyInstalled)
                     GameTypeEnabled = false;
@@ -215,13 +215,13 @@ namespace Wabbajack
 
         private void UpdateFiltersSettings()
         {
-            if (!settings.isPersistent)
+            if (!settings.IsPersistent)
                 return;
             if (!string.IsNullOrEmpty(GameType))
                 settings.Game = GameType;
             if (Search != null)
                 settings.Search = Search;
-            settings.showNSFW = ShowNSFW;
+            settings.ShowNSFW = ShowNSFW;
             settings.OnlyInstalled = OnlyInstalled;
         }
     }
--- a/Wabbajack/View Models/Gallery/ModListGalleryVM.cs
+++ b/Wabbajack/View Models/Gallery/ModListGalleryVM.cs
@@ -73,6 +73,10 @@ namespace Wabbajack
                 if (OnlyInstalled)
                     GameTypeEnabled = false;
                 Search = settings.Search;
+                // subscribe to save signal
+                MWVM.Settings.SaveSignal
+                    .Subscribe(_ => UpdateFiltersSettings())
+                    .DisposeWith(this.CompositeDisposable);
             }
             else
                 GameType = ALL_GAME_TYPE;
@@ -85,7 +89,6 @@ namespace Wabbajack
                     Search = string.Empty;
                     GameType = ALL_GAME_TYPE;
                     GameTypeEnabled = true;
-                    UpdateFiltersSettings();
                 });
 
             var random = new Random();
@@ -141,7 +144,6 @@ namespace Wabbajack
                 .Filter(this.WhenAny(x => x.OnlyInstalled)
                     .Select<bool, Func<ModListMetadataVM, bool>>(onlyInstalled => (vm) =>
                     {
-                        UpdateFiltersSettings();
                         if (!onlyInstalled) return true;
                         if (!GameRegistry.Games.TryGetValue(vm.Metadata.Game, out var gameMeta)) return false;
                         return gameMeta.IsInstalled;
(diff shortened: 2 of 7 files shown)


Wabbajack/Settings.cs
```diff
@@ -93,6 +94,18 @@ public class CompilerSettings
...
+    [JsonObject(MemberSerialization.OptOut)]
+    public class FiltersSettings : ViewModel
+    {
+        public bool showNSFW { get; set; }
+        public bool OnlyInstalled { get; set; }
+        public string Game { get; set; }
+        public string Search { get; set; }
+        private bool _isPersistent = true;
+        public bool isPersistent { get => _isPersistent; set => RaiseAndSetIfChanged(ref _isPersistent, value); }
+    }
```
Noggog, line 108: `showNSFW` should be `ShowNSFW`, as it's public.  Same with `IsPersistent`

LordOfRhun: Hey nice catch, i am new with github is there any way i can attach a commit to this review or simply commiting again is fine ?

Noggog: If you make another commit with the changes and just push, it'll sync and work out

LordOfRhun: ok thanks !

Noggog:
Looking peachy from a "it works" perspective!  8)  I think there's a few things to tweak or discuss a little bit.

One is the pattern you're using for syncing the settings, namely the `UpdateFiltersSettings` call.  I don't think that should be called everywhere like it is now.

So far we've got two patterns in the WJ app for saving settings:
- Settings are applied to the VMs at startup, and saved from the VMs to the settings objects/file at shutdown.  So the in either direction are only done once during the lifetime of the program; Changes to values in the VM members mid-run are not synced every time to the settings file objects. 
- Views bind straight to the settings object directly, with no middleman members on the VM.  As such, any change to a view control by the user is directly modified in the settings object via bindings, and vice versa.  This is only done in one place so far: the performance settings.  Can't remember the exact reasoning offhand for needing this.  I'd have to brush up again.

I think you can get away with 1) and just do a sync on shutdown.  To detect when that's happening, you'd maybe write something like this:
```
MWVM.Settings.SaveSignal
    .Subscribe(_ =>UpdateFiltersSettings())
    .DisposeWith(this.CompositeDisposable);
```
The settings file has an observable for when the shutdown is happening, so you just latch on and do the sync once then

LordOfRhun:
Yes you are right, there is no reason to always update the settings on runtime, i will be making some changes to save them only on application exit !

Noggog:
Second thing is the `GameTypeEnabled` member.  

First off, I'm unsure why `Only Installed` is incompatible with `Selected Game`?  Shouldn't those filters be theoretically able to combine?  Not the most useful filter combination, but I don't know that we need to go out of our way to block the intersection.

If we do, though, I think we can still improve the way `GameTypeEnabled` is implemented.

One feature I think you're not taking full advantage of is Rx (Reactive Extensions), and the fact that all our ViewModel's properties are subscribable, thanks to their [Reactive] attribute which powers via magic it under the hood.

So, rather than chasing down all the ways `OnlyInstalled` can change, we can just subscribe to it directly:
```
this.WhenAny(x => x.OnlyInstalled)
    .Subscribe(val=>GameTypeEnabled = val)
    .DisposeWith(CompositeDisposable);
```
This way, no matter how OnlyInstalled is changed, it'll update GameTypeEnabled properly.

------

But.. we can even go further.  GameTypeEnabled is really just an inverse of OnlyInstalled and it's 100% a derivative value of OnlyInstalled.  If OnlyInstalled is true, GameTypeEnabled must be false, etc.  We can modify the above code to instead make GameTypeEnabled a derivative, readonly property that's "driven" by OnlyInstalled.
```
// My derivative property's declaration in the class.  Notice it's now readonly
private readonly ObservableAsPropertyHelper<bool> _GameTypeEnabled;
public bool GameTypeEnabled => _GameTypeEnabled.Value;

...

// In the ctor, I can then wire it up to be driven by OnlyInstalled 
_GameTypeEnabled = this.WhenAny(x => x.OnlyInstalled)
    .Transform(val=>!val)
    .ToGuiProperty(this, nameof(GameTypeEnabled ));
```
Now, GameTypeEnabled will always be the inverse of OnlyInstalled.  AND, there is no possible way that could ever be otherwise, as no one can even set GameTypeEnabled to be something different, as it's now a derivative readonly property.

------

But I think all that is overkill anyway, and mostly just to show off some patterns that you'll see elsewhere in WJ and might be helpful in the future for you.   What you really want to do is just a inverse logic directly in the binding.

Just swap your XAML binding to:
```
IsEnabled="{Binding OnlyInstalled, Converter={StaticResource InverseBooleanConverter}}"
```
Boom, done.  You can find the code for `InverseBooleanConverter` in the same WJ project, to see the gritty specifics.

Can then just delete `GameTypeEnabled` outright; No need to declare a new variable that needs syncing with OnlyInstalled.

Wabbajack/View Models/Gallery/ModListGalleryVM.cs
```diff
@@ -102,7 +123,13 @@ public ModListGalleryVM(MainWindowVM mainWindowVM)
 
             OnlyInstalledCheckedCommand = ReactiveCommand.Create(() =>
             {
-                GameType = ALL_GAME_TYPE;
+                if (OnlyInstalled)
+                {
+                    GameType = ALL_GAME_TYPE;
```
Noggog, line 128: I think GameType could also be "reset" via Rx instead of making an `OnlyInstalledCheckedCommand` member.  Just do an Rx subscription to OnlyInstalled as shown above, and reset GameType to All at that point.

Noggog:
Other than that, looking good!  Thanks for taking your time to add stuff to WJ. 8)
Much appreciated!

LordOfRhun:
In my understanding OnlyInstalled is to sort only the modlists for the games you have installed. I disabled the selectedGame because you could enable OnlyInstalled and then select a game you dont have installed and get not results. you could also select a game you do have installed so having OnlyInstalled enabled was making so sense to me. One solution would be to have OnlyInstalled in the comboBox below All :)

LordOfRhun:
> Other than that, looking good!  Thanks for taking your time to add stuff to WJ. 8)
> Much appreciated!

I found this application and community very nice and helpfull. I like to program and saw the opportunity to help you guys while getting better at WPF, a framework i do not have much experience in. Really it's my pleasure !

Noggog:
>  because you could enable OnlyInstalled and then select a game you dont have installed and get not result

Yeah, I think in any filtering situation, it's always a possibility to filter too narrowly and get no results.  It's not something to be too concerned about, I don't think.  

> I like to program and saw the opportunity to help you guys while getting better at WPF, a framework i do not have much experience in.

Awesome!  Well feel free to ask any questions you have along the way in discord.  I always like talking theory and bouncing off ideas and whatnot. 8)

LordOfRhun:
Thanks for the great explanations, Rx seems very useful. I made the requested changes.

Noggog:
Looking good!

halgari:
@Noggog are you good with the state of this then?

Noggog:
I have not cloned and run it myself, just looked on GitHub.  

I see a few minor refactors I'll probably suggest now looking again.  Will clone once those are done

Wabbajack/View Models/Gallery/ModListGalleryVM.cs
```diff
@@ -24,12 +24,12 @@ public class ModListGalleryVM : BackNavigatingVM
 
         public ObservableCollectionExtended<ModListMetadataVM> ModLists { get; } = new ObservableCollectionExtended<ModListMetadataVM>();
 
+        private FiltersSettings settings;
```
Noggog, line 27: I think we can remove this member.  The filter settings are always available via walking the links to the settings object (`MWVM.Settings.FilterSettings`).  Having a second variable here means there's two places the same information is stored, which has a low (but unnecessary) potential of desync.  If we store the filter settings object here, but then later someone goes and creates a new filter settings object on the "official" settings object in MWVM.. well now the two systems are pointing/using different filter settings objects.

Wabbajack/View Models/Gallery/ModListGalleryVM.cs
```diff
@@ -56,8 +56,23 @@ public ModListGalleryVM(MainWindowVM mainWindowVM)
...
+            // load persistent filter settings
+            settings = MWVM.Settings.Filters;
+            if (settings.IsPersistent)
+            {
+                GameType = !string.IsNullOrEmpty(settings.Game) ? settings.Game : ALL_GAME_TYPE;
+                ShowNSFW = settings.ShowNSFW;
+                OnlyInstalled = settings.OnlyInstalled;
+                Search = settings.Search;
+                // subscribe to save signal
+                MWVM.Settings.SaveSignal
```
Noggog, line 69: This subscription needs to not be inside the `IsPersistent` if statement.  Consider if the settings on-disk have persistence turned off on startup.  Now this constructor runs, and the subscription/callback to save to the settings won't be made at all, even if IsPersistent is turned on mid-run.

Wabbajack/View Models/Gallery/ModListGalleryVM.cs
Noggog: If you wanted to have a member here for convenience, so that you didn't have to write `MWVM.Settings.FilterSettings` every time, then you can utilize C#'s properties system to make a "derivative" property that just "reroutes" to the same variable.  That way there's not actually a second variable that needs to be synced with.
```
private FiltersSettings settings => MWVM.Settings.FilterSetting;
```

Wabbajack/View Models/Gallery/ModListGalleryVM.cs
```diff
@@ -173,11 +196,20 @@ public override void Unload()
...
-            {
-                gameEntries.Add(EnumExtensions.GetDescription<Game>(gameType));
-            }
+            gameEntries.AddRange(EnumExtensions.GetAllItems<Game>().Select(gameType => gameType.GetDescription<Game>()));
             return gameEntries;
         }
+
+        private void UpdateFiltersSettings()
+        {
+            if (!settings.IsPersistent)
```
Noggog, line 205: I think there is a subtle edge case here.  Consider that the user has had IsPersistent on, and has their filters saved.  Now they start WJ up again, and decide they don't like that.  They turn IsPersistent off.  Now when this function gets called to save the settings, we short circuit and return.

Well, the end result is that his previously saved filter settings are just left as-is, so the next time they start up the filters won't be clean, they'll have the stale "last saved" filters.

This call needs to actively wipe the filter settings clean if IsPersistent is off

Wabbajack/View Models/Gallery/ModListGalleryVM.cs
```diff
@@ -173,11 +196,20 @@ public override void Unload()
...
+                return;
+            if (!string.IsNullOrEmpty(GameType))
+                settings.Game = GameType;
+            if (Search != null)
+                settings.Search = Search;
```
Noggog, line 210: Same with these lines, right now if the current input is null, then the settings will be left with whatever they had in them from before.

But I think rather than worrying too much, I think we can just remove the if statements here?  I don't believe we need to care if things are null and weed those from being saved.

LordOfRhun:
made the requested adjustments :)

Noggog:
I have clicked ALL the buttons, and they seem to work. 8)

PR #825, opened 2020-05-11
Update azure-pipelines.yml for Azure Pipelines

halgari:


erri120:
👀 code coverage? We already use [Codecov](https://codecov.io/) for the website, are you also going to add that to this project?

halgari:
I was hoping to, but the tests take a serious runtime hit. I'll look into this again when I've sped up the tests a bit

Issue #630, opened 2020-03-11
Displaying progress in the task bar


erri120:
![](https://docs.microsoft.com/en-us/dotnet/media/wpfshell-taskbariteminfo.png?view=netframework-4.8)
(talking about the overlay)
We already have an internal progress bar so maybe we can easily implement this.

References:
[TaskbarItemInfo.ProgressState Property](https://docs.microsoft.com/en-us/dotnet/api/system.windows.shell.taskbariteminfo.progressstate?view=netframework-4.8)
[Microsoft.WindowsAPICodePack-Shell](https://www.nuget.org/packages/Microsoft.WindowsAPICodePack-Shell/)

PR #905, opened 2020-06-02
Initial Steps toward Implementing TaskbarItemInfo

j9ac9k:
This pull request is to implement #630 Displaying Progress in the Taskbar using TaskbarItemInfo

The functionality does not work at present (hence the Draft PR state).  I haven't done any C# development, but started this PR early in case I was missing something obvious.

Thanks.

### DIFF
--- a/Wabbajack/View Models/MainWindowVM.cs
+++ b/Wabbajack/View Models/MainWindowVM.cs
@@ -34,6 +34,10 @@ namespace Wabbajack
         [Reactive]
         public ViewModel ActivePane { get; private set; }
 
+        private readonly ObservableAsPropertyHelper<Percent> _percentCompleted;
+
+        public Double PercentCompleted => (double)_percentCompleted.Value; 
+
         public ObservableCollectionExtended<IStatusMessage> Log { get; } = new ObservableCollectionExtended<IStatusMessage>();
 
         public readonly Lazy<CompilerVM> Compiler;
@@ -144,7 +148,15 @@ namespace Wabbajack
 
             OpenTerminalCommand = ReactiveCommand.Create(() => OpenTerminal());
             
-
+            _percentCompleted = Observable.CombineLatest(
+                this.WhenAny(x => x.Installer.Value.PercentCompleted),
+                this.WhenAny(x => x.Compiler.Value.PercentCompleted),
+                (Percent installer, Percent compiler) =>
+                {
+                    // add logic to take te active percentage, installer or compiler
+                    return installer;
+                })
+                .ToGuiProperty(this, nameof(PercentCompleted));
         }
 
         private void OpenTerminal()
--- a/Wabbajack/Views/MainWindow.xaml
+++ b/Wabbajack/Views/MainWindow.xaml
@@ -81,4 +81,11 @@
             </Button>
         </mahapps:WindowCommands>
     </mahapps:MetroWindow.RightWindowCommands>
+    <Window.TaskbarItemInfo>
+        <TaskbarItemInfo x:Name="taskBarItemInfo1" 
+                     ThumbnailClipMargin="80,0,80,140"
+                     Description="Progress"
+                     ProgressValue="{Binding PercentCompleted}">
+        </TaskbarItemInfo>
+    </Window.TaskbarItemInfo>
 </mahapps:MetroWindow>


Wabbajack/View Models/MainWindowVM.cs
```diff
@@ -34,6 +34,10 @@ public class MainWindowVM : ViewModel
         [Reactive]
         public ViewModel ActivePane { get; private set; }
 
+        private readonly ObservableAsPropertyHelper<Percent> _percentCompleted;
+
+        public Double PercentCompleted => (double)_percentCompleted.Value;
```
Noggog, line 39: I would remove the cast from Percent to Double.
Despite being two lines, they should be considered one field.  What type is this field?  Double, or Percent?

You can make it double if you want, in which case you need to convert from Percent to double in the construction.  (somewhere before the ToGuiProperty call)

Wabbajack/View Models/MainWindowVM.cs
```diff
@@ -144,7 +148,15 @@ public MainWindowVM(MainWindow mainWindow, MainSettings settings)
...
             OpenTerminalCommand = ReactiveCommand.Create(() => OpenTerminal());
             
-
+            _percentCompleted = Observable.CombineLatest(
+                this.WhenAny(x => x.Installer.Value.PercentCompleted),
+                this.WhenAny(x => x.Compiler.Value.PercentCompleted),
+                (Percent installer, Percent compiler) =>
+                {
+                    // add logic to take te active percentage, installer or compiler
+                    return installer;
```
Noggog, line 157: You didn't sub in any logic for selecting the proper percent to use.  Right now it's always going to show the installer's percent, and will never make use of the compiler's percent.

You might be wondering how to tell which to choose/use:
 - Can either check and return whichever one is not 0 or 1, as a hacky but probably decently accurate way.
 - Can add WhenAny calls and retrieve CompilerVM's `Compiling` property, and the similar one from InstallerVM.  By adding them into the properties that will trigger a refresh, you'll also need to add them as parameters to the lambda for use.  Now the lambda knows which side is the active one, so it'll know which percentage to return for sure.

Wabbajack/Views/MainWindow.xaml
```diff
@@ -81,4 +81,11 @@
             </Button>
         </mahapps:WindowCommands>
     </mahapps:MetroWindow.RightWindowCommands>
+    <Window.TaskbarItemInfo>
+        <TaskbarItemInfo x:Name="taskBarItemInfo1" 
+                     ThumbnailClipMargin="80,0,80,140"
+                     Description="Progress"
+                     ProgressValue="{Binding PercentCompleted}">
```
Noggog, line 88: So the concepts I mentioned about XAML binding here to MainWindowVM's PercentCompleted seems like it won't work for TaskbarItemInfo, specifically.  I tried it out and it doesn't seem to take.  Seems to be an oddball legacy Winforms control that doesn't support binding?  Either that, or I messed something up.

Either way, it for sure is an oddball control:  It doesn't appear on the control tree of any live XAML inspector or anything.. which makes sense as it's it's own oddball thing:  the taskbar widget/item.

TLDR, i don't think XAML bindings will work.. so this is one of the few times codebehind systems like you were initially using are actually appropriate and required due to it being older tech.

I would suggest a callback something like this inside MainWindow.xaml.cs:
```
 _mwvm.WhenAny(x => x.PercentCompleted)
    .Subscribe(p =>
    {
        this.taskBarItemInfo1.ProgressValue = (double)p;
    });
```

This would likely need to go inside the `Initialized` lambda, as taskbarItemInfo1 isn't initialized until that callback occurs.

Wabbajack/Views/MainWindow.xaml
```diff
@@ -81,4 +81,11 @@
             </Button>
         </mahapps:WindowCommands>
     </mahapps:MetroWindow.RightWindowCommands>
+    <Window.TaskbarItemInfo>
+        <TaskbarItemInfo x:Name="taskBarItemInfo1" 
+                     ThumbnailClipMargin="80,0,80,140"
+                     Description="Progress"
```
Noggog, line 87: I think you need to set the "ProgressState" variable.  Default will not have any progress mechanics active

halgari:
closed due to inactivity

PR #933, opened 2020-06-26
BSA Prep

Noggog:
As minimal a PR to get Wabbajack.Common, BSA, and VFS to .net standard.

Eventually we may feel like trimming the library BSA relies on the be leaner with less WJ concepts, but for now this is fine.


Noggog:
Threw in some other stuff:
- Comments on the libraries enabled so they show up for end users
- cleaned up editorconfig
- Noticed some missing awaits

PR #959, opened 2020-07-14
Work without build server

erri120:
Does a simple check to see if the build server is alive and caches that value. Also fixed the problem where the main window does not appear after the splash screen. Apparently was an issue with `Metrics.Send("started_wabbajack", fvi.FileVersion)`. Wrapping that in `Task.Run(() => ...)` so it doesn't block solved the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ### Changelog
 
+#### Version - next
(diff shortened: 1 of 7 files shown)


tr4wzified:
Awesome, I like that you included install / downloads directories in the logs too, should've probably been included a long time ago

PR #977, opened 2020-07-26
Install config work

Noggog:
- Added error icon triangle under install play button if there are blocking problems.
- Added tooltip styling to limit width to 500.
- Adjusted error text for MO2Installer unexpected files.
- Added filepicker error glow
- Added a global XAML styler file.  If XAML styler installed, it will add a more consistent formatting.
- Resaved all XAML with the rules


Noggog:
image

PR #1010, opened 2020-08-06
Handle extracting exes

halgari:


halgari:
No longer needed, and this approach won't work for our needs

PR #1024, opened 2020-08-11
Bsa Optimizations

Noggog:
Main goal was to get a BSAReader.TryGetFolder call which would allow searching for strings quicker/minimal.

- BSAReader refactored to make most operations very lazy and deferred.
- Added BSAReader.TryGetFolder
- Optimizations for async file I/O.  Mainly within AbsolutePath where they get constructed.
   - Async flag enabled
   - Buffer increased to 1MB.  Less queries into the file the better with this pattern
   - https://stackoverflow.com/questions/39353119/filestream-readasync-very-slow-compared-to-read
- BSAReader split to several files instead of one
- RelativePath now has a skip validation option in ctor.  BSAs use it internally
- MemorySlice ported over.  It's a Span-style struct that can live on the heap.  C#'s stock Memory is slower, for some reason, which is why I created it
- Marked PDB files to be included in Common/BSA libraries
- Removed RxUI from Wabbajack.Common.  Had to move over UserInterventions to Lib, but I think that should be fine?

Like I mentioned in discord, this refactor seems to have made BSAs 10-15x faster at least even for operations that are fully iterating all files, and even better for selective queries using the TryGetFolder

I still had trouble running the unit tests.  I did push this through Mutagen's passthrough systems, and my few little Synthesis test patchers, so that flexed it a decent amount.  Pretty major refactor the class, though, so might be a bug hiding


Noggog:
Here's my timing test.  Nothing crazy
```
Stopwatch sw = new Stopwatch();
sw.Start();
for (int i = 0; i < 10000; i++)
{
    var bsa = BSAReader.Load(new AbsolutePath(@"D:\Games\steamapps\common\Skyrim Special Edition\Data\Skyrim - Textures3.bsa"));
    var files = bsa.Files.ToArray();
    var names = files.Select(f => f.Path.ToString()).ToArray();
    if (i % 100 == 0)
    {
        System.Console.WriteLine(i);
    }
}
sw.Stop();
System.Console.WriteLine($"{sw.ElapsedMilliseconds}ms");
System.Console.ReadLine();
```

Noggog:
Should be noted, the speed tests are more about the iteration of the file headers within the BSA.  Code for extracting the file data to a stream is mostly untouched except for enabling the async flags.  Should get a little boost from that, but didn't test explicitly.

Also, the other BSA types were untouched.  Will probably get to them at some point with similar changes

PR #1073, opened 2020-09-05
mark WJ installed archives as installed in MO2.

EzioTheDeadPoet:
A simple change to the the way .meta files get handled by Wabbajack to mark archives used by Wabbajack as installed when launching ModOrganizer2.


erri120:
resolves #1073

EzioTheDeadPoet:
Doesn't it need to be  `return return $"{string.Join("\n", GetMetaIni())}\ninstalled=true";` since string.join() uses the `\n` as a separator to divide the previous output ?

PR #1104, opened 2020-09-22
Optional Logging

Noggog:
Some more Wabbajack.Common sprucing to make it public friendly.  Was previously logging internals to file without any control levers from the library user's perspective.

- Moved WJ local app data filewatching system to a lazy setup.  (Skip for normal library users)
- Moved logging functionality to separate file for organization (still part of Utils class)
- Added LoggingSettings.LogToFile.  This defaults to off for ease of use for the library users.  I added lines to turn it on for both the GUI and the server.

Library users should hopefully now be able to access Utils.LogMessages to listen to WJ events and funnel appropriately

### DIFF
--- a/Wabbajack.Common/Logging.cs
+++ b/Wabbajack.Common/Logging.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reactive.Subjects;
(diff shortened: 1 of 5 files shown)


Noggog:
I don't think the unit tests failures are legit.   Was able to run some of the failing ones locally

halgari (approved):

PR #1249, opened 2021-01-09
Filter downloads for file size before hashing archives.

Unnoen:
This PR makes Wabbajack check the sizes of files to see if they match what is required before hashing them.

### DIFF
--- a/Wabbajack.Lib/AInstaller.cs
+++ b/Wabbajack.Lib/AInstaller.cs
@@ -279,12 +279,41 @@ namespace Wabbajack.Lib
         public async Task HashArchives()
         {
             Utils.Log("Looking for files to hash");
-            var toHash = DownloadFolder.EnumerateFiles()
+            // Enumerate all files.
+            var allFiles = DownloadFolder.EnumerateFiles()
                 .Concat(Game.GameLocation().EnumerateFiles())
                 .ToList();
+
+            // Create a Dictionary with filesize as the key.
+            var hashDict = new Dictionary<long, List<AbsolutePath>>();
+
+            // Populate the Dictionary with all files we found, using a list for the paths to allow for duplicated filesizes.
+            allFiles.ForEach(o =>
+            {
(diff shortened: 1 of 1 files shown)


Wabbajack.Lib/AInstaller.cs
```diff
@@ -279,12 +279,41 @@ public async Task<bool> DownloadArchive(Archive archive, bool download, Absolute
...
+                if (hashDict.ContainsKey(o.Size))
+                {
+                    hashDict[o.Size].Add(o);
+                }
+                else
+                {
+                    hashDict.Add(o.Size, new List<AbsolutePath> { o });
+                }
+            });
+
```
halgari, line 302: This can just be:

```csharp
var hashDict = allFiles.GroupBy(f => f.Size).ToDictionary(group => group.Key);
```

Wabbajack.Lib/AInstaller.cs
```diff
@@ -279,12 +279,41 @@ public async Task<bool> DownloadArchive(Archive archive, bool download, Absolute
...
+            // Create a list for the files we filter.
+            var toHash = new List<AbsolutePath>();
+
+            // Iterate through required archives, adding any files we have that match filesize.
+            ModList.Archives.ForEach(o => {
+                if (hashDict.ContainsKey(o.Size))
+                {
+                    hashDict[o.Size].ForEach(o => toHash.Add(o));
+                }
+            });
```
halgari, line 312: This can be: 

```csharp
var toHash = ModListArchives.Where(a => hashDict.ContainsKey(a.Size).SelectMany(a => hashDict[a.Size]).ToList();

```

Wabbajack.Lib/AInstaller.cs
Unnoen: Man, I really need to learn all the things Linq allows, this massively simplifies it, thanks.

PR #1254, opened 2021-01-10
Fix detection of Windows 10 versions

LostDragonist:
I'm honestly a little skeptical of this.  My logs are indeed reporting my version as `10.0.19042.0` which makes sense for version 20H2.  The main question I have is why was it checking 6.2 before and why WJ seems to think I have a different version of Windows now.

### DIFF
--- a/Wabbajack/Views/MainWindow.xaml.cs
+++ b/Wabbajack/Views/MainWindow.xaml.cs
@@ -42,7 +42,8 @@ namespace Wabbajack
 
                 Utils.Log($"Detected Windows Version: {p.WindowsVersion}");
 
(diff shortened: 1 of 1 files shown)


LostDragonist:
Here's some info.  https://docs.microsoft.com/en-us/windows/win32/sysinfo/operating-system-version.
> For applications that have been manifested for Windows 8.1 or Windows 10. Applications not manifested for Windows 8.1 or Windows 10 will return the Windows 8 OS version value (6.2). To manifest your applications for Windows 8.1 or Windows 10, refer to Targeting your application for Windows.

Given that...  I guess I changed it to check for Windows 8.1 or Windows 10?  Or some weird amalgamation as the logic is a bit strange.

PR #1259, opened 2021-01-11
Log pagefile size and warn if disabled or suboptimal.

Unnoen:
I've set the recommended to 20GB (20000) as that seems to handle anything thrown at it.


Wabbajack/Util/SystemParametersConstructor.cs
```diff
@@ -56,6 +56,7 @@ public static SystemParameters Create()
                 ScreenHeight = height,
                 VideoMemorySize = video_memory,
                 SystemMemorySize = (long)memory.ullTotalPhys,
+                SystemPageSize = (long)memory.ullTotalPageFile - (long)memory.ullTotalPhys
```
Unnoen, line 59: Reported pagefile includes physical, I've just subtracted that so we're left with the true amount.

PR #1284, opened 2021-01-26
Using MO2Name instead of MO2ArchiveName for GameFileSourceDownloader

erri120:


erri120:
From discord (https://discord.com/channels/605449136870916175/622649396428931088/803639831480762378):
> WJ can't find a fitting downloader for the game files because it thinks those files are for Skyrim SE and not Skyrim VR.
The reason being that we use the `MO2ArchiveName` when re-creating the meta ini for those files: https://github.com/wabbajack-tools/wabbajack/blob/master/Wabbajack.Lib/Downloaders/GameFileSourceDownloader.cs#L99
The problem is that `MO2ArchiveName` for `SkyrimVR` is `skyrimse`:
https://github.com/wabbajack-tools/wabbajack/blob/master/Wabbajack.Common/GameMetaData.cs#L405
Thus the `GameFileSourceDownloader` will grab `SkyrimSE` from the registry instead of `SkyrimVR`: https://github.com/wabbajack-tools/wabbajack/blob/master/Wabbajack.Lib/Downloaders/GameFileSourceDownloader.cs#L21

PR #1369, opened 2021-03-14
adds option to include save files if wanted.

### LINKED ISSUE #1356: Ability to set a tag to skip the "skipping" of savefiles

Some people want to include preset saves for their lists.

A tag-file in the root folder that forces the list compiler to skip the step intended to exclude the save files, so the save files get included.

just uploading them somewhere.

**Additional context**

mostly a note for me so I don't forget to do this eventually.

EzioTheDeadPoet:
- Closes  #1356 
Mainly to allow a different form of MCM automation.


erri120:
I think it would be better to just change https://github.com/wabbajack-tools/wabbajack/blob/master/Wabbajack.Lib/CompilationSteps/IgnoreSaveFiles.cs

EzioTheDeadPoet:
fair point

EzioTheDeadPoet:
Doing a local test rn of an adjusted version.
Any clue why the master is failing and therefore my PRs based on it ? 
Cause GitHub says that the build while it builds just fine locally.

erri120:
> 
> 
> Doing a local test rn of an adjusted version.
> Any clue why the master is failing and therefore my PRs based on it ?
> Cause GitHub says that the build while it builds just fine locally.

some tests fail

EzioTheDeadPoet:
So compile with the "tagfile" set and install worked. testing without it now.

EzioTheDeadPoet:
Only merge shortly before release, since I added the instructions for this to the `README.md`

PR #1370, opened 2021-03-14
include mo2 splash screen if present

### LINKED ISSUE #1365: Detect and inline custom MO2 splashscreen
File is named `splash.png` and is located in the root MO2 directory. Modlists authors currently have to archive this file, upload it and create a meta for it which is kinda overkill.

EzioTheDeadPoet:
+ closes #1365


halgari:
Pretty sure we can use the "IncludeStartsWith" step instead and not have to add another class

EzioTheDeadPoet:
Works for me tbh I somehow totally missed that.

PR #1394, opened 2021-03-26
Add GameFinder lib

erri120:
Looks ugly because I still have to add an OriginHandler to the GameFinder lib.


erri120:
Updated GameFinder to 1.5.0 for no throw, tests should work now.

PR #1457, opened 2021-05-24
Logging Quality of Life updates

Unnoen:
This PR focuses on making the log files generated by Wabbajack easier to debug [read: prettier].
I've set this PR as a Draft to garner feedback and reviews.

There are now multiple logging levels, each with it's own function instead of one function with a level parameter.
\- `Utils.Trace("msg");` instead of `Utils.Log(Level.Trace, "msg");`
(The Info level function has remained `Utils.Log()` to reduce unnecessary refactoring)

The functions utilise `System.Runtime.CompilerService` to receive the CallerFilePath, which we then parse to get the calling file.

There are five levels currently specified:
- `Trace` receives a string and logs to file only.
- `Info` receives a string and logs to file and display. (Can specify to not write to an output.)
- `Warn` similar to Info.
- `Error` has two functions, one receives a string, the other an exception. Otherwise similar to Info.
- `Fatal` receives an exception, will throw after logging. (Can specify to not throw.)

I've done my best to accurately label logging with the correct levels, but there may be room for improvement and/or changes. There's quite a few files to go through but it should be pretty easy to cruise through since it's mostly renaming a single function. (`Utils.Log` -> `Utils.Error`, etc)

Other changes:
1. Log the time the program was started.
2. Removed the output string from 7zipReturnError as it was redundant and unused. (The StdErr stream is always logged.) and added the drive name to the description.
3. `MO2Installer.cs` now prints the failed download archive URL alongside the name.
4. Added additional error logging to `EpicGameStoreHandler.cs`, `DownloadDispatcher.cs`, `HTTPDownloader.cs`

### DIFF
--- a/Wabbajack.Common/CircuitBreaker/WithAutoRetry.cs
+++ b/Wabbajack.Common/CircuitBreaker/WithAutoRetry.cs
@@ -27,7 +27,7 @@ namespace Wabbajack.Common
                 retries += 1;
                 if (retries > maxRetries)
                     throw;
-                Utils.Log($"(Retry {retries} of {maxRetries}), got exception {ex.Message}, waiting {delay!.Value.TotalMilliseconds}ms");
+                Utils.Warn($"(Retry {retries} of {maxRetries}), got exception {ex.Message}, waiting {delay!.Value.TotalMilliseconds}ms");
                 await Task.Delay(delay.Value);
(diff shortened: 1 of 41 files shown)


Unnoen:
Example log in VSCode: 
image

halgari:
we can look at this again on the new codebase

PR #1483, opened 2021-06-11
Switching from SharpDX.DXGI to Silk.NET.DXGI

erri120:
There is currently a bug in Silk.NET which leads to some funky values so we have to wait until https://github.com/dotnet/Silk.NET/issues/521 is fixed


erri120:
Issue was fixed in Silk.NET only have to wait for the next package update.

erri120:
Silk.NET updated to 2.6.0 and fixed the issue.

PR #1527, opened 2021-07-02
Fix LL Attachment Metadata check

ForgottenGlory:
I believe the issue with the LL attachments is that the DownloaderTests' metadata check do not account for an alternative INI formatted for attachments. This should, if I'm understanding correctly, fix that.


halgari:
Fixed in master

PR #1575, opened 2021-08-14
Add Nehrim support

LostDragonist:


Wabbajack.VirtualFileSystem/VirtualFile.cs
```diff
@@ -266,10 +266,10 @@ private IndexedVirtualFile ToIndexedVirtualFile()
             {
                 return self;
             }
-            catch (Exception)
+            catch (Exception e)
             {
-                Utils.Log($"Error while examining the contents of {relPath.FileName}");
-                throw;
+                Utils.Log($"Error while examining the contents of {relPath.FileName}: {e.Message}");
+                return self;
```
halgari, line 272: We want this to re-throw, otherwise the VFS result could be corrupt

halgari:
needs to be done on the new codebase

PR #1835, opened 2022-01-21
Blazor Update

erri120:


Wabbajack.App.Blazor/App.xaml.cs
```diff
@@ -33,24 +34,18 @@ private static IServiceCollection ConfigureServices(IServiceCollection services)
         services.AddAllSingleton<ILoggerProvider, LoggerProvider>();
         services.AddTransient<MainWindow>();
         services.AddSingleton<SystemParametersConstructor>();
-        services.AddSingleton<GlobalState>();
+        services.AddSingleton(typeof(IStateContainer), typeof(StateContainer));
```
halgari, line 37: can be done without `typeof(`

`services.AddSingleton<IStateContainer, StateContainer>();`

PR #1847, opened 2022-01-27
Blazor updates and minor changes to other internal projects.

Unnoen:
The Installing process has been split into three pages, Select, Configure and Installing.

Added [Blazored.Modal](https://github.com/Blazored/Modal) and [Blazored.Toast](https://github.com/Blazored/Toast) for notifications and overlays with some placeholder components. These are yet to be styled.

The Installer and Compiler StatusUpdates now also include a StatusCategory. This is mostly to add another thing we can use to make the design of the Blazor pages a bit fancier, and gives a more general idea of where the process is up to.

I added a CommentRegex to IniParser's configuration. It's not in the object initialiser due to IniParser overwriting it, so we have to set it *after* that.
This fixes SSE Display Tweak's comments causing an exception.
It also now logs the remapped files.

SystemParametersConstructor is now called in the App.xaml.cs constructor, as it won't log the assembly information until it's first initialised, and it's kind of important to have.

NLog now archives logs on startup, rather than appending to the same file. The ui memory target only contains the message content to avoid clutter.
The file logging layout has also been changed, I've changed it to something I feel is good, but this is subjective and can be altered if need be.


Wabbajack.App.Blazor/App.xaml.cs
```diff
@@ -22,6 +24,7 @@ public App()
             .ConfigureServices(services => ConfigureServices(services))
             .Build()
             .Services;
+        _serviceProvider.GetRequiredService<SystemParametersConstructor>();
```
erri120, line 27: Is this some leftover?

Wabbajack.App.Blazor/Components/ProgressBar.razor
```diff
@@ -1,18 +1,34 @@
...
+            .Sample(TimeSpan.FromMilliseconds(250))
+            .DistinctUntilChanged()
+            .Subscribe(p => {
+                CurrentProgress = p.Value;
+                if (textPercentage) Text = p.ToString();
+                InvokeAsync(StateHasChanged);
+            });
+    
+        return base.OnInitializedAsync();
+    }
```
erri120, line 33: No need to call `base.OnInitializedAsync()`, also don't use the *Async method if you don't have async code. Also why the `string.IsNullOrEmpty(Text)` check? Isn't this always true?

Wabbajack.App.Blazor/Pages/Gallery.razor.cs
```diff
@@ -44,55 +46,50 @@ protected override async Task OnInitializedAsync()
...
-                };
 
-                InvokeAsync(StateHasChanged);
-                // Dispatcher.CreateDefault().InvokeAsync(StateHasChanged);
-            }, () => { StateContainer.TaskBarState = new TaskBarState(); });
+            DownloadProgress = progress;
+
+            var dispose = progress
+                .Sample(TimeSpan.FromMilliseconds(250))
+                .Subscribe(p => {
```
erri120, line 82: Would still slap a `.DistinctUntilChanged(p => p.Value)` in here

Wabbajack.App.Blazor/Pages/Install/Installing.razor.cs
```diff
@@ -26,84 +27,36 @@ public partial class Configure
...
-            if (downloadPath is not null) DownloadPath = (AbsolutePath)downloadPath;
-        }
-        catch (Exception e)
-        {
-            Logger.LogError(e, "Exception selecting download folder");
-        }
-    }
-
-    private async Task Install()
+    private async void Install()
```
erri120, line 56: Never use `async void`, when doing async/await always return `Task`, `Task<T>` or the `ValueTask` equivilant. Also change `void OnInitialized()` back to `async Task OnInitializedAsync()`.

Wabbajack.App.Blazor/wwwroot/index.html
```diff
@@ -32,12 +34,19 @@
 <div id="app"></div>
 
 <script src="_framework/blazor.webview.js"></script>
+<script src="_content/Blazored.Modal/blazored.modal.js"></script>
 <script>
     async function setImageUsingStreaming(imageElementId, imageStream) {
         const arrayBuffer = await imageStream.arrayBuffer();
         const blob = new Blob([arrayBuffer]);
         document.getElementById(imageElementId).src = URL.createObjectURL(blob);
     }
```
erri120, line 43: Are we still using this somewhere? I saw `getBlobUrlFromStream` so this can probably be deleted.

Wabbajack.Compiler/ACompiler.cs
```diff
@@ -195,7 +197,7 @@ internal async Task<(RelativePath, AbsolutePath)> IncludeString(string str)
     public async Task<bool> GatherMetaData()
     {
         _logger.LogInformation("Getting meta data for {count} archives", SelectedArchives.Count);
-        NextStep("Gathering Metadata", SelectedArchives.Count);
+        NextStep("Building", "Gathering Metadata", SelectedArchives.Count);
```
erri120, line 200: Should probably make the categories into constants

Wabbajack.App.Blazor/Pages/Install/Configure.razor.cs
```diff
@@ -0,0 +1,131 @@
...
+        catch (Exception e)
+        {
+            Logger.LogWarning(e, "Exception loading previous settings for {Name}", ModlistPath);
+        }
+
+        try
+        {
+            var imageStream = await StandardInstaller.ModListImageStream(ModlistPath);
+            var dotnetImageStream = new DotNetStreamReference(imageStream);
+            StateContainer.ModlistImage = new string(await JSRuntime.InvokeAsync<string>("getBlobUrlFromStream", dotnetImageStream));
```
erri120, line 84: why the `new string`? `JSRuntime.InvokeAsync<string>` should return `ValueTask<string>`.

Wabbajack.App.Blazor/State/IStateContainer.cs
```diff
@@ -17,9 +17,18 @@ public interface IStateContainer
...
+    AbsolutePath InstallPath { get; set; }
+    
+    IObservable<AbsolutePath> DownloadPathObservable { get; }
+    AbsolutePath DownloadPath { get; set; }
+    
     IObservable<ModList?> ModlistObservable { get; }
     ModList? Modlist { get; set; }
     
+    IObservable<string> ModlistImageObservable { get; }
+    string ModlistImage { get; set; }
```
erri120, line 30: Should be nullable. Would require changes elsewhere and some `is not null` checks but it provides more safety.

Wabbajack.App.Blazor/Components/StepLogger.razor
```diff
@@ -0,0 +1,20 @@
...
+
+@code {
+    [Parameter]
+    public int ShownSteps { get; set; }
+    [Parameter]
+    public bool Reverse { get; set; }
+
+    [CascadingParameter]
+    public Installing value { get; set; }
+}
```
erri120, line 20: I saw the `<CascadingValue Value=this>` below, I would just remove this component and inline this inside the `Installing` page. There is no reason to have this as a component, it just degrates performance. This component can't even be re-used because you used `[CascadingParamtere] public Installing value { get; set; }` so just inlined this entire thing.

Wabbajack.App.Blazor/Components/ProgressBar.razor
```diff
@@ -1,18 +1,34 @@
...
+    [Parameter] public IObservable<Percent> ProgressObserver { get; set; }
 
+    [Parameter] public string Text { get; set; }
+
+    protected override Task OnInitializedAsync()
+    {
+        var textPercentage = string.IsNullOrEmpty(Text);
+        ProgressObserver
+            .Sample(TimeSpan.FromMilliseconds(250))
+            .DistinctUntilChanged()
```
erri120, line 25: `DistinctUntilChanged(p => p.Value)`

Wabbajack.App.Blazor/Main.razor
```diff
@@ -1,13 +1,14 @@
...
-<Router AppAssembly="@GetType().Assembly">
-    <Found Context="routeData">
-        <RouteView RouteData="@routeData" DefaultLayout="@typeof(MainLayout)"/>
-    </Found>
-    <NotFound>
-        <h1>Not found</h1>
-        <p>Sorry, there's nothing here.</p>
-    </NotFound>
-</Router>
+<CascadingBlazoredModal>
```
erri120, line 4: I've checked the code for Blazored.Modal and they use `CascadingValue` everywhere instead of just using DI. Using `CascadingParameter` is even more expensive than normal `Parameter` so just inject `IModalService`.

Wabbajack.App.Blazor/Pages/Gallery.razor.cs
```diff
@@ -22,13 +23,14 @@ public partial class Gallery
     [Inject] private NavigationManager NavigationManager { get; set; } = default!;
     [Inject] private ModListDownloadMaintainer Maintainer { get; set; } = default!;
 
-    private Percent DownloadProgress { get; set; } = Percent.Zero;
+    [CascadingParameter] public IModalService Modal { get; set; }
```
erri120, line 26: Use `[Inject] private IModalService Modal { get; set; } = default!;`

Wabbajack.App.Blazor/App.xaml.cs
```diff
@@ -30,30 +33,43 @@ private static void SetupLogging(ILoggingBuilder loggingBuilder)
 
         var fileTarget = new FileTarget("file")
         {
-            FileName = "log.log"
+            FileName = "logs/Wabbajack.current.log",
+            ArchiveFileName = "logs/Wabbajack.{##}.log",
+            ArchiveOldFileOnStartup = true,
+            MaxArchiveFiles = 10,
+            Layout = "${processtime} [${level:uppercase=true}] (${logger}) ${message:withexception=true}",
```
erri120, line 40: Continuing our heated discussion from the Discord: pipes are nice `"${processtime}|${level:uppercase=true}|${logger}|${message:withexception=true}"` which is just the default NLog layout but `processtime` instead of `longdate`.

erri120 (changes requested):

Wabbajack.App.Blazor/App.xaml.cs
Unnoen: This is so the assembly information (build, entry point, etc) actually gets logged at the program launch. There's probably a better way but this works so it's what I went with.

Wabbajack.App.Blazor/Components/ProgressBar.razor
Unnoen: `Text` is null if the parameter isn't passed, I'll add the `?` to the type to make it clearer.

Wabbajack.App.Blazor/Pages/Install/Installing.razor.cs
Unnoen: Changing `void OnInitialized()` to `async Task OnInitializedAsync()` makes the install process block UI updates.

Wabbajack.Compiler/ACompiler.cs
Unnoen: Yeah, I'm not too sure where to put them however.

Wabbajack.App.Blazor/Pages/Install/Configure.razor.cs
Unnoen: I can't remember, fixed now.

Wabbajack.App.Blazor/Main.razor
Unnoen: The CascadingBlazoredModal is required here otherwise the modal won't render. I don't think we need to be this pedantic over performance as we're not doing anything even remotely close to something where UI performance matters this much.

Wabbajack.App.Blazor/App.xaml.cs
Unnoen: 😠

PR #2078, opened 2022-09-20
Make NLog print all exception information.

Unnoen:
Currently, NLog's `${message:withexception=true}` renderer skips important parts like file and line numbers.

This adds the exception renderer and prints all details of exceptions.


snakefoot:
Notice both `${exception:tostring}` and `${message:withexception=true}` outputs the result of `Excetion.ToString()`

But  `Excetion.ToString()` only provide line-numbers and source-code-filenames when PDB-/symbol-files are available in the deployment folder.

PR #2082, opened 2022-09-24
Set PublishTrimmed in buildall.bat instead of PublishedTrim

barcharcraz:
PublishTrimmed is (I suspect) what you really want, the CLI sets this property in it's csproj (with the correct name). This will probably make binaries smaller!

I discovered this by chance when building all the projects as non-self-contained apps (since I know I already have the .net framework).

### DIFF
--- a/buildall.bat
+++ b/buildall.bat
@@ -7,8 +7,8 @@ mkdir c:\tmp\publish-wj
 
 dotnet clean
 dotnet restore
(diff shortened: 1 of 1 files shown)


halgari:
Thanks for the report I fixed up the launcher app, for the main app, trimming broke a few things so I removed the switch alltogether

PR #2147, opened 2022-10-31
Add Karryn's Prison game support

madtisa:
Added Karryn's Prison game support to be able to post modlists for this game


erri120:
Very based 👍

PR #2162, opened 2022-11-02
#1958 witcher enhanced edition support

EzioTheDeadPoet:
closes #1958.

PR #2163, opened 2022-11-02
#2159 skyrim gog detection metadata

### LINKED ISSUE #2159: GOG detection of Skyrim Special Edition + Anniversary Upgrade [Pending]
**What's the problem?**

@Althro: Wants people to be able to install his list with GOG versions of the game.

**The solution @Althro would like:**

Adding detection info for the GOG version of the Game.

**Alternatives I have considered:**

Telling people to wait or buy the steam version or wait. 

**Reasons why I personally would wait with adding support:**

Adding it now could likely cause an influx of people that only own a GOG copy of Skyrim trying to install a majority of list that don't support GOG causing a support nightmare for associated Support Discord servers and the main Wabbajack Discord server.

EzioTheDeadPoet:
Closes #2159 as requested by @Althro since his list already is adjusted to support Steam and GOG at the same time.


EzioTheDeadPoet:
same reason for a second merge from within the repository as #2162

PR #2169, opened 2022-11-04
Game support readme update

EzioTheDeadPoet:
- add Dragon's Dogma Dark Arisen GOGID
- update game list in readme


EzioTheDeadPoet:
@halgari Ideally merge this just before the next update releases, since otherwise the README.md is showing DDDA GOG support when it isn't there yet. My mistake because I did mix the change for it into the branch that I intended only for the readme, But for writing the readme I had the code open which is where I noticed it and had to do it. 😅

erri120:
I think we should sort the table by game name.

EzioTheDeadPoet:
I will look into changing that then 👍

PR #2173, opened 2022-11-05
Alert when running as admin

erri120:


thx4nothing:
Why is this a thing? I always ran it as admin (I disabled UAC on my computer so it automatically opens as admin) and never had any problems. Now I cannot use Wabbajack anymore (without workarounds or building it myself). If there is no good reason other than "sAfeTy", could we reverse this change? Or make a command line argument that lets us disable this check. I mean, what does wabbajack care if I start it as admin or not. It should be the decision of the user, even if wabbajack doesn't need admin rights.

erri120:
> Why is this a thing? I always ran it as admin (I disabled UAC on my computer so it automatically opens as admin) and never had any problems. Now I cannot use Wabbajack anymore (without workarounds or building it myself). If there is no good reason other than "sAfeTy", could we reverse this change? Or make a command line argument that lets us disable this check. I mean, what does wabbajack care if I start it as admin or not. It should be the decision of the user, even if wabbajack doesn't need admin rights.

1) "sAfeTy"
2) some users run Wabbajack as admin when they encounter an issue. They _think_ it will be fixed by running it as admin when in reality the issue lies somewhere else
3) running Wabbajack as admin can result in undefined behavior like working directory being the system32 folder and since WJ has admin rights it can and will fuck with files in the working dir

EzioTheDeadPoet:
Wabbajack is a powerful tool for doing critical file operations in batch.
this includes:

- downloading
- extracting
- copying
- moving

and the most critical:

- deleting 

files.

Considering that running as admin allows it to have free reign over any file and can cause it to operate in system32 (as erri explained) on some systems makes it actually a valid and long overdue security feature especially considering how many people get thought when debugging gaming stuff "just run it as administrator".

thx4nothing:
> > Why is this a thing? I always ran it as admin (I disabled UAC on my computer so it automatically opens as admin) and never had any problems. Now I cannot use Wabbajack anymore (without workarounds or building it myself). If there is no good reason other than "sAfeTy", could we reverse this change? Or make a command line argument that lets us disable this check. I mean, what does wabbajack care if I start it as admin or not. It should be the decision of the user, even if wabbajack doesn't need admin rights.
> 
>     1. "sAfeTy"
> 
>     2. some users run Wabbajack as admin when they encounter an issue. They _think_ it will be fixed by running it as admin when in reality the issue lies somewhere else
> 
>     3. running Wabbajack as admin can result in undefined behavior like working directory being the system32 folder and since WJ has admin rights it can and will fuck with files in the working dir

I understand your reasons. From a typical user perspective, I think this is fair. That's why I would propose to solve it like I said, as a command line argument like "--disable-admin-check". This way, the typical user can benefit from the enhanced safety and users like me (I saw some on discord too) can still use wabbajack.

In my case, I can't even run it as a limited user through process explorer (it then tries to use the system32 folder). I would have to enable UAC, restart my pc, do wabbajack stuff, disable UAC and restart again. I build wabbajack with the admin check removed and it worked perfectly again.

If the mentioned solution would be acceptable, I could work on it and create a pull request.

Thanks for taking your time to respond to my comment.

thx4nothing:
> Wabbajack is a powerful tool for doing critical file operations in batch. this includes:
> 
>     * downloading
> 
>     * extracting
> 
>     * copying
> 
>     * moving
> 
> 
> and the most critical:
> 
>     * deleting
> 
> 
> files.
> 
> Considering that running as admin allows it to have free reign over any file and can cause it to operate in system32 (as erri explained) on some systems makes it actually a valid and long overdue security feature especially considering how many people get thought when debugging gaming stuff "just run it as administrator".

In my case, it has the opposite effect. As I mentioned in my other comment, it uses system32 when running as a limited user. When running as my normal admin account, it never uses system32.

thx4nothing:
> Why do you have UAC disabled?

I don't think it should matter in this discussion what I (and other users) choose to do with my computer and I won't change my ways just because one software suddenly prohibits me to use it. Other software I use never implemented this kind of admin check (afaik).

But anyways, I choose to disable it a few years ago because I found it too restricting and had some problems with accessing files, namely in the Program Files folder.

While I acknowledge that it would not be a wise decision for the average user to disable it, I never had any problems with UAC disabled, except for this wabbajack update and one old game.

EzioTheDeadPoet:
We won't change this because you think it is an inconvenience. Especially since the way windows handles disabling UAC is inherently flawed AF. And simply by enabling a backdoor for this change any "just run it as administrator"-kid will just do the additional steps and mess with it potentially breaking their shit.

EzioTheDeadPoet:
But feel free to try the latest update to Wabbajack because it likely fixed ur issue anyways already.

thx4nothing:
> We won't change this because you think it is an inconvenience. Especially since the way windows handles disabling UAC is inherently flawed AF. And simply by enabling a backdoor for this change any "just run it as administrator"-kid will just do the additional steps and mess with it potentially breaking their shit.

That's unfortunate to hear...
Thanks for taking your time to answer my question anyway.


> But feel free to try the latest update to Wabbajack because it likely fixed ur issue anyways already.

Unfortunately, it didn't fix it.

erri120:
If you really don't want to change your UAC settings, you can just batch script that uses [RUNAS](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-R2-and-2012/cc771525(v=ws.11)) to change your trust level:

```
RUNAS /trustlevel:<level> Wabbajack
```

Espionage724:
> Considering that running as admin allows it to have free reign over any file and can cause it to operate in system32 (as erri explained) on some systems makes it actually a valid and long overdue security feature especially considering how many people get thought when debugging gaming stuff "just run it as administrator".

With all due respect, Wabbajack isn't my anti-virus. I want to download mods. Why does Wabbajack feel the need to try to protect me from some theoretical scenario? Even if _somehow_ something goes wrong strictly because of Wabbajack being ran as-admin, so what? I'm a big boy and take responsibility for running a "non-standard" set-up.

This silly implementation wasted my time trying to workaround it and I was not happy about having to create a different account. If it was reasonable to download modpacks without Wabbajack, I wouldn't be complaining about this. The above runas command from @erri120 works nicely though:

```
RUNAS /trustlevel:0x20000 Wabbajack.exe
```

PR #2198, opened 2022-11-10
Simplified #2195

erri120:


EzioTheDeadPoet:
This doesn't work they way my change was intended to work!!!!

This method missed some files in the past, which is exactly why I made it a separate function.

erri120:
> This doesn't work they way my change was intended to work!!!!
> 
> This method missed some files in the past, which is exactly why I made it a separate function.

Which files? The only difference I can see between `InstallIncludedDownloadMetas` and `SetDownloadMetasToHideInMO2` is the fact that `SetDownloadMetasToHideInMO2` will create metas for every file in the downloads folder while `InstallIncludedDownloadMetas` will only create metas for files that are going to be installed and leaves existing files as is.

EzioTheDeadPoet:
This doesn't guarantee that every file is hidden. Especially when installing into an existing load order. And hiding EVERY file was the whole deal behind this.

And I made sure that the meta data for existing files stays the way it is except that they get hidden.

But I just realized I need to account for the fact that some files might already have `removed=false` in them which I don't know how MO2 would handle that if I just append a `removed=true` after that.

But that's a bugfix I will make later today when needed after work.

But as it stands it was intended to also Hide non installed files while keeping their other meta data and installed status the way WJ and MO2 treat it now. (do power users can unhide and delete not-installed archives since they are likely files that got corrupted during download etc.)

erri120:
> This doesn't guarantee that every file is hidden. Especially when installing into an existing load order. And hiding EVERY file was the whole deal behind this.

It guarantees that every file to be installed gets hidden. If you start with no existing MO2 instance and install a Wabbajack modlist then all installed files will be hidden. Then, when you update the modlist, all new files will also be hidden. Of course, if the user added custom files, those will not be hidden since the user installed them.

EzioTheDeadPoet:
> > This doesn't guarantee that every file is hidden. Especially when installing into an existing load order. And hiding EVERY file was the whole deal behind this.
> 
> It guarantees that every file to be installed gets hidden. If you start with no existing MO2 instance and install a Wabbajack modlist then all installed files will be hidden. Then, when you update the modlist, all new files will also be hidden. Of course, if the user added custom files, those will not be hidden since the user installed them.

This doesn't fix the user error of people already using shared downloads folders and trying to install mods they find in there. And therefore doesn't fix the issue that initiated that change. And another issue with the current method is that if people had to restart an installation previously failed(not matching the hash) downloads won't be hidden as well and users frequently ask questions about those and try to install them. It might not always break stuff, but it is an issue and user interaction than can be avoided by hiding everything by default.

erri120:
> > > This doesn't guarantee that every file is hidden. Especially when installing into an existing load order. And hiding EVERY file was the whole deal behind this.
> > 
> > 
> > It guarantees that every file to be installed gets hidden. If you start with no existing MO2 instance and install a Wabbajack modlist then all installed files will be hidden. Then, when you update the modlist, all new files will also be hidden. Of course, if the user added custom files, those will not be hidden since the user installed them.
> 
> This doesn't fix the user error of people already using shared downloads folders and trying to install mods they find in there. And therefore doesn't fix the issue that initiated that change. And another issue with the current method is that if people had to restart an installation previously failed(not matching the hash) downloads won't be hidden as well and users frequently ask questions about those and try to install them. It might not always break stuff, but it is an issue and user interaction than can be avoided by hiding everything by default.

If user interaction is that wild, then it's fine, but iterating twice over every downloaded file is not great. I'll update `InstallIncludedDownloadMetas` so we just have one place where we write meta files.

EzioTheDeadPoet:
The reason being that this way the `installed=true` marks the correct files but still hides them and everything else.

erri120:
I've updated the code:

- if the file is part of the modlist:
  - add `installed=true`
- if the file is not part of the modlist:
  1) check if `removed` is already set
  2) if not, add `removed=true`

EzioTheDeadPoet:
Works for me 👍

EzioTheDeadPoet:
Closes #2199.

PR #2205, opened 2022-11-13
Fix writing metas task

EzioTheDeadPoet:
should be pushed out ASAP because right now installations seem to be in an infinite loop for at least one user without the fix.
[Wabbajack.current (1).log](https://github.com/wabbajack-tools/wabbajack/files/9996990/Wabbajack.current.1.log)

### DIFF
--- a/Wabbajack.Installer/StandardInstaller.cs
+++ b/Wabbajack.Installer/StandardInstaller.cs
@@ -229,9 +229,9 @@ public class StandardInstaller : AInstaller<StandardInstaller>
 
         _logger.LogInformation("Writing Metas");
         await _configuration.Downloads.EnumerateFiles()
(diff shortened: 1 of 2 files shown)


erri120 (approved):

EzioTheDeadPoet:
Oh the tests are on the wrong .NET again.

Wabbajack.Installer/StandardInstaller.cs
```diff
@@ -231,6 +231,7 @@ private async Task WriteMetaFiles(CancellationToken token)
         await _configuration.Downloads.EnumerateFiles()
```
erri120, line 231: Now that I think about it, this should probably be a `.Where(download => download.Extension == Ext.Meta)` for increased performance.

erri120 (changes requested):

PR #2280, opened 2023-01-21
Add Mount & Blade II: Bannerlord support?

Maelstrom8:
So, with the MO2 plugin that Holt59 created, I was able to make a modlist with M&B Bannerlord in MO2, and I'd like to share it through WJ, but it's not on the UI. Would be great..

I don't actively use github and so I just looked at how someone did it for Karryn's Prison - I didn't do any commits for the readme because of the wiki being a thing. If anything is done incorrectly, feel free to tell me.


EzioTheDeadPoet:
Is the MO2 plugin already part of the main MO2 build? If so the genericMO2Game needs to be set to true.

For the game u used as reference it is only false because the person wanting support for it didn't make it clear if they were using a MO2 plugin for the game.

Besides that does Bannerlord have GOG support?

In that case we could add the ID itself.

Maelstrom8:
I've set that to true. Also, yes - Bannerlord is also for GOG, EPG and the Microsoft Store. Too many platforms :S.

I just Steam to mod it.

PR #2340, opened 2023-05-21
Fix issues introduced by resumable downloads

UrbanCMC:
This should fix the downloader issues introduced with the resumable downloads feature.

I've also included some of my logging changes, because they were useful in getting a clearer picture of where and how downloads may fail.

**Notable changes**:
* Downloads now reserve their full size on disk when starting
-- Fixes high RAM usage
* Limited download threads to a max of 8, regardless of available threads
-- Fixes hanging when starting a large amount of downloads simultaneously
* If a download fails, it will be retried 3 times in the `SingleThreadedDownloader` before the caller can perform its own error handling (e.g. Use a different NexusMods server)
-- Can fix failing downloads on flaky connections
* The size of download jobs is dynamically increased (by 10% of current size) if the downloader gets more than the expected number of bytes. This appears to be caused by the downloader sometimes having to re-download chunks if they don't match.
-- Fixes downloads appearing stuck in job view.

**Known issues**:
v3.1.0.0 appears to have introduced an issue where the whole UI freezes after starting the installation process, while continuing to work in the background. From user reports, limiting the number of simultaneous downloads to 8 does not seem to fix the issue.

I wasn't able to find any cause for this, and users were able to fix this by either waiting for some time, or completely restarting WJ.

### DIFF
--- a/Wabbajack.Downloaders.Dispatcher/DownloadDispatcher.cs
+++ b/Wabbajack.Downloaders.Dispatcher/DownloadDispatcher.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
+using Wabbajack.Common;
 using Wabbajack.Downloaders.Interfaces;
 using Wabbajack.Downloaders.VerificationCache;
 using Wabbajack.DTOs;
@@ -50,7 +51,9 @@ public class DownloadDispatcher
 
         using var downloadScope = _logger.BeginScope("Downloading {Name}", a.Name);
         using var job = await _limiter.Begin("Downloading " + a.Name, a.Size, token);
-        return await Download(a, dest, job, token, proxy);
+        var hash = await Download(a, dest, job, token, proxy);
+        _logger.LogInformation("Finished downloading {name}. Hash: {hash}; Size: {size}/{expectedSize}", a.Name, hash, dest.Size().ToFileSizeString(), a.Size.ToFileSizeString());
+        return hash;
     }
 
     public async Task<Archive> MaybeProxy(Archive a, CancellationToken token)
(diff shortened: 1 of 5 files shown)


EzioTheDeadPoet:
@halgari 
Best you look into it since you mentioned stuff about the thread settings not being recognized correctly on discord and maybe finding a better solution than a flat out cap to a maximum of 8 since imo that would be a big downgrade for people used to downloading with more threads.
Also I am expecting this request to fail the tests again since that has happened in the past with all pull requests coming from forked repositories.

UrbanCMC:
Just out of curiosity, are most of the servers used for downloading actually bandwidth limited? I applied the cap under the impression that most of them use whatever you have available, so capping to 8 downloads wouldn't have much effect on the throughput.
If this is a concern, I would expect raising the cap to 12 or even 16 downloads wouldn't be too risky. But I'm hesitant about going higher than that.

EzioTheDeadPoet:
Nexus isn't throttled if you have premium, but other sites sometimes have less performance than peoples ISPs and network cards could achieve.

But for people doing the installation without premium it is either with 1MB/s or now 2MB/s and 3MB/s if they have the supporter account status which people get after a previous premium subscription has run out.

So for people doing it manually having all available threads useable would be preferable.

UrbanCMC:
So, how do you want to proceed on this?

Should I just remove the limit on download threads, or would you prefer we update the current debug version and wait to see if things get worse again?

EzioTheDeadPoet:
Maybe make it a release with a different tag on your repo. 
That way the workaround still links to a tested version, but the new testing version I would link in hte bleeding-edge testing channel. 
And I would also ping Modlist Devs to encourage their users that have high core & thread count CPUs to test that version if they think they can handle navigating to an extracted programm and executing it.
That way we would get some testing done and the old links and pins still link to a know to be working version.

UrbanCMC:
Debug version is uploaded here: https://github.com/UrbanCMC/wabbajack/releases/tag/3.1.0-debug-no-thread-limit

The WJ log will contain one of the following two lines when an installation is started, to identify which version is running:
`[DEBUG] (Wabbajack.Installer.StandardInstaller) ResumableDownloader Debug v4`
or
`[DEBUG] (Wabbajack.Installer.StandardInstaller) ResumableDownloader Debug v4 - No download thread limit`

EzioTheDeadPoet:
> The only change from that version is the removal of the download cap to 8 threads. This version will attempt to start a new download for each thread available on the user's CPU.

Does this mean it is ignoring the config file of WJ for allowed threads ? because always using that would probably be the best option!
That way we could make an inital config based on the users specs in regards to RAM, and CPU Threads and they could still tweak it if they like to!

EzioTheDeadPoet:
Once you can confirm that it uses the settings file I'll push the ping to the people on discord.

UrbanCMC:
> Does this mean it is ignoring the config file of WJ for allowed threads ? because always using that would probably be the best option!
No, that part works the same as it did before my change.

The only change I did was basically `Min(8, configuredThreads)` at runtime, specifically for download threads.
So everyting else still happens the way it used to:
* When the config is first created, it saves the number of available threads as the max for each resource.
* Manually changing the settings file allows using more/fewer resource threads than the number of CPU threads.

IMO this should be ready for testing.

PR #2355, opened 2023-07-11
Disable opening readme website when WJ file is opened, keep it to only open on install complete

### LINKED ISSUE #2279: Doubled Readme Page Opening
**Describe the bug**

When you begin installation of a modlist, the readme page is opened twice - two tabs are opened for the same page. On finishing installation, the readme page is opened only once. The latter seems correct and the former should only open the readme page once.

**Additional Context**

Windows 11

**Wabbajack Version**

3.0.0.5

**Logs**

image

JanuarySnow:
Fixes #2279 

This may be down to preference as per the linked issue, but the argument that when testing a failing install, having the webpage open every time being annoying, has some merit.

This removes the code that opens the website when a WJ file is first opened, and only opens it when it is most relevant - when the install is complete and things need to be read.

There may be situations where having the readme open to troubleshoot install issues may be more friendly to the user, but ill leave that to others to decide, this "Fix" is just here if you want it.


JanuarySnow:
Linked issue was originally referencing an actual bug where it opened twice at first, and then once more at completion, the later comments in that issue moved onto debating whether it should open at the start at all, and thats what I acted upon in error

PR #2370, opened 2023-07-16
Commenting out LoversLab login from the settings page to avoid user confusion.

EzioTheDeadPoet:
The current way to implement LL downloads is `manualURL=` + `prompt=` https://wiki.wabbajack.org/modlist_author_documentation/Meta%20Files.html#manualurl

The branch contains Vector Plexus in its name since I thought it was broken due to no one using it, but it was already updated to the vectorplexis domain change and seems to be working so far.

### DIFF
--- a/Wabbajack.App.Wpf/App.xaml.cs
+++ b/Wabbajack.App.Wpf/App.xaml.cs
@@ -182,7 +182,9 @@ namespace Wabbajack
             services.AddTransient<LoversLabLoginHandler>();
 
             // Login Managers
-            services.AddAllSingleton<INeedsLogin, LoversLabLoginManager>();
+            
+            //Disabled because they are currently not used and bro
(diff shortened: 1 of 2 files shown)


Wabbajack.App.Wpf/App.xaml.cs
```diff
@@ -182,7 +182,9 @@ private static IServiceCollection ConfigureServices(IServiceCollection services)
             services.AddTransient<LoversLabLoginHandler>();
 
             // Login Managers
-            services.AddAllSingleton<INeedsLogin, LoversLabLoginManager>();
+            
+            //Disabled because they are currently not used and bro
```
JanuarySnow, line 186: and bro... :P

EzioTheDeadPoet: *bro-ken
🤣

PR #2374, opened 2023-07-19
Additional UI changes to help user confusion and information

### LINKED ISSUE #2372: User placed the Installation folder *inside* the downloads' folder. Meta chaos and failed installation occurred.
**Describe the bug**

As per title. - from Wildlander Discord.

**Additional Context**

I told you that they would find a "unique" way to bypass your checks @JanuarySnow 

**Wabbajack Version**

Wabbajack Version: 3.2.0.0

**Logs**
[message(5).txt](https://github.com/wabbajack-tools/wabbajack/files/12089516/message.5.txt)

JanuarySnow:
Fixes #2372 

Added a few more features to the UI to help ease support burden.

Added check that the install folder isnt inside the download folder

Removed the overwrite installation checkbox - it didnt save its state anymore as far as I can tell, and clicking it saved you one click by avoiding the "are you sure you want to overwrite?" prompt.  - so clicking once to avoid one click isnt much of a good deal, and it caused user confusion as to its actual purpose.

I've added a hard drive space checker to the install and download paths, ( if same drive itll add the required download and install together ), and itll block the install if there isnt enough space

Ive added some red warning text under the file picker textbox, to additionally outline the current blocking error, so user does not need to be told "hover your mouse over warning label" to find out what the current error is.

Ive expanded the error message text for existing files in the chosen install path, to make it clear that if user is updating an existing modlist, then it is fine to continue.

Updated error message text for protected folders to be more clear about what a protected folder means, and what user might need to do to avoid targeting them

Ive added an additional button to the install failure screen, that links to the wabbajack wiki, so users might press it on an install failure and get some self-help

I have more comprehensive plans in this area, but I see this as more pressing in the meantime due to the number of support queries regarding these install blocking errors.

( self-conscious disclaimer - I am obviously still very new to ReactiveUI/WPF, and c# in general , coming from other languages I may not be aware of some common patterns, so apologies for the roughness of this  - currently nobody else is stepping up to fix these things in the short-term, so I will, but I will gladly step back if somebody implements these things better )

Pictures here: 

image

image

### DIFF
--- a/Wabbajack.App.Wpf/View Models/Installers/InstallerVM.cs
+++ b/Wabbajack.App.Wpf/View Models/Installers/InstallerVM.cs
@@ -336,7 +336,7 @@ public class InstallerVM : BackNavigatingVM, IBackNavigatingVM, ICpuStatusVM
 
         if (KnownFolders.IsInSpecialFolder(installPath) || KnownFolders.IsInSpecialFolder(downloadPath))
         {
-            yield return ErrorResponse.Fail("Can't install a modlist into Windows protected locations - such as Downloads,Documents etc, please make a new folder for the modlist.");
+            yield return ErrorResponse.Fail("Can't install into Windows locations such as Documents etc, please make a new folder for the modlist - C:\Wabbajack\ for example.");
         }
 
         if (installPath.ToString().Length > 0 && downloadPath.ToString().Length > 0 && !HasEnoughSpace(installPath, downloadPath)){
--- a/Wabbajack.App.Wpf/View Models/Installers/InstallerVM.cs
+++ b/Wabbajack.App.Wpf/View Models/Installers/InstallerVM.cs
@@ -336,7 +336,7 @@ public class InstallerVM : BackNavigatingVM, IBackNavigatingVM, ICpuStatusVM
 
         if (KnownFolders.IsInSpecialFolder(installPath) || KnownFolders.IsInSpecialFolder(downloadPath))
         {
-            yield return ErrorResponse.Fail("Can't install into Windows locations such as Documents etc, please make a new folder for the modlist - C:\Wabbajack\ for example.");
+            yield return ErrorResponse.Fail("Can't install into Windows locations such as Documents etc, please make a new folder for the modlist - C:\\ModList\\ for example.");
         }
 
         if (installPath.ToString().Length > 0 && downloadPath.ToString().Length > 0 && !HasEnoughSpace(installPath, downloadPath)){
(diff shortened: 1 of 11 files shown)


EzioTheDeadPoet (approved):

Wabbajack.App.Wpf/View Models/Installers/InstallerVM.cs
```diff
@@ -325,10 +338,43 @@ private IEnumerable<ErrorResponse> Validate()
 
         if (KnownFolders.IsInSpecialFolder(installPath) || KnownFolders.IsInSpecialFolder(downloadPath))
         {
-            yield return ErrorResponse.Fail("Can't install a modlist into Windows protected locations - such as Downloads, Documents etc");
+            yield return ErrorResponse.Fail("Can't install a modlist into Windows protected locations - such as Downloads,Documents etc, please make a new folder for the modlist.");
```
Aegis270, line 339: This may not be explicit enough, since users may just make a new folder inside a protected location since that technically fulfills the requirements laid out here (Not directly inside a protected folder, and in a new folder). 

I'd recommend adding a suggestion as already done for the launch path.

Wabbajack.App.Wpf/View Models/Installers/InstallerVM.cs
```diff
@@ -336,7 +336,7 @@ private IEnumerable<ErrorResponse> Validate()
 
         if (KnownFolders.IsInSpecialFolder(installPath) || KnownFolders.IsInSpecialFolder(downloadPath))
         {
-            yield return ErrorResponse.Fail("Can't install a modlist into Windows protected locations - such as Downloads,Documents etc, please make a new folder for the modlist.");
+            yield return ErrorResponse.Fail("Can't install into Windows locations such as Documents etc, please make a new folder for the modlist - C:\Wabbajack\ for example.");
```
EzioTheDeadPoet, line 339: **THIS IS A BAD EXAMPLE !!**

Because with this users will wipe the recommended WJ location making WJ try to delete itself...

😅

JanuarySnow: Oops I forgot what i was looking at, I thoguth for a moment this was the wabbajack.exe location error! sorry my bad

EzioTheDeadPoet: not a big deal, just happy I caught it 😂

JanuarySnow:
Added some code to revert the popup window warning about files existing in install path, as it was unwieldy and fired multiple times whenever validate() was called.

Instead just added it as an error return type alongside all the other error types, so it prints the warning on the screen, and ive reintroduced the "overwrite installation" checkbox to make that warning go away and re-enable the play button.

EzioTheDeadPoet:
Duplicate of #2416

Technically the other one is the duplicate but I am trying to user Github trigger words to make it more fun. 😉

PR #2376, opened 2023-07-29
Add optional JSON flag to only show version number for modlist image in install view

### LINKED ISSUE #1524: Option to display the version number instead of the modlist name

My modlist already has its name in the banner. Having the name empty or the version number instead breaks the metrics.

having an option to hide the name when compiling and potentially replacing it with the version number if needed would be ideal.

Doing nothing? But it has the modlist name twice then.

**Additional context**

Current:
image

Ideal:
Wabbajack_TwCVlxeKSi

JanuarySnow:
Fixes #1524 

Bit of an obscure one, but it dosnt impact those who dont want it.

Add an optional flag in the modlist metadata json for `DisplayVersionOnlyInInstallerView = true` 

if this is true, then the version number only is displayed over the modlist image in the installer view, to remain consistent with the same behavior in the gallery tile view ( where there is no overlaid title for images that have title in them )

This new behavior is down to personal preference even more so dosnt follow the same flag, and uses a new one.

image

### DIFF
--- a/Wabbajack.App.Wpf/View Models/Gallery/ModListMetadataVM.cs
+++ b/Wabbajack.App.Wpf/View Models/Gallery/ModListMetadataVM.cs
@@ -83,6 +83,13 @@ namespace Wabbajack
         public string VersionText { get; private set; }
 
         [Reactive]
+        public bool ImageContainsTitle { get; private set; }
+
+        [Reactive]
+
+        public bool DisplayVersionOnlyInInstallerView { get; private set; }
+
+        [Reactive]
         public IErrorResponse Error { get; private set; }
 
         private readonly ObservableAsPropertyHelper<BitmapImage> _Image;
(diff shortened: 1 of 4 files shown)


tr4wzified:
Looks fine to me, but I wouldn't stray too far from the C# coding guidelines that are used elsewhere in the project. Makes it a bit of a mess in the code having snake_case and PascalCase mixed in there combined with { and } being on the same lines as the statement

Wabbajack.App.Wpf/View Models/Gallery/ModListMetadataVM.cs
```diff
@@ -82,6 +82,13 @@ public class ModListMetadataVM : ViewModel
         [Reactive]
         public string VersionText { get; private set; }
 
+        [Reactive]
+        public bool ImageContainsTitle { get; private set; }
+
+        [Reactive]
+
+        public bool display_version_only_in_installer_view { get; private set; }
```
erri120, line 90: Please use PascalCase.

Wabbajack.DTOs/ModList/ModListMetadata.cs
```diff
@@ -30,6 +30,8 @@ public class ModlistMetadata
     [JsonPropertyName("image_contains_title")]
     public bool ImageContainsTitle { get; set; }
 
+    [JsonPropertyName("display_version_only_in_installer_view")] public bool display_version_only_in_installer_view { get; set; }
```
erri120, line 33: Same here, use PascalCase.

erri120 (changes requested):

JanuarySnow:
> Looks fine to me, but I wouldn't stray too far from the C# coding guidelines that are used elsewhere in the project. Makes it a bit of a mess in the code having snake_case and PascalCase mixed in there combined with { and } being on the same lines as the statement

Right you are, sorry, muscle memory kicks in and I forget about c# stuff :D

erri120 (approved):

tr4wzified (approved):

PR #2432, opened 2023-10-25
Actually use the compile --outputPath option if the user provides one

majcosta:
It was just not being used, defaulting to the parent of installPath. it still does, if the user does not specify a directory using --outputPath


EzioTheDeadPoet:
I will test it in the preview branch and see to include it in the changes I have planned to work on today.

PR #2471, opened 2024-01-04
update Final Fantasy VII: RI metadata

EzioTheDeadPoet:
Fix an issue where the main executable was set to a "launcher" executable and not the actual main executable that contains the proper version number. (This is important for the hashes to target the correct game version)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ### Changelog
 
+#### Version - TBD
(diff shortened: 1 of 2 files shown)


EzioTheDeadPoet:
@halgari the latest test seems to fail due to a rate limiter on our end or nexus end. But should be unrelated to the changes I made.

tr4wzified (approved):

tr4wzified:
Added as part of 3.4.2.0

PR #2524, opened 2024-04-04
Add Fallout New Vegas from Epic Games support

TDarkShadow:
At the moment it works with just Fallout New Vegas set to English.
However, Wabbajack doesn't set the Mod Organizer 2 game path correctly so the user gets a prompt to browse to their installed Epic Games' Fallout New Vegas installation.


EzioTheDeadPoet:
@halgari CI is failing on the nexus login 😅

Issue #2587, opened 2024-06-12
Starfield support added, but compilation currently impossible due to file format differences


cacophony-wj:
**Describe the bug**

Although Starfield and its hashes were added to the repository, WJ currently fails to recognize some of the files during compilation, presumably due to changes in the files' structure

**Additional Context**

Starfield, Windows 11

**Wabbajack Version**

3.6.1.1

Attached as a file / link

[Wabbajack.00.log](https://github.com/user-attachments/files/15810645/Wabbajack.00.log)

This report is made on behalf of another author, the log is not from my machine

tr4wzified:
Got a beta release out with Starfield support.

- Source: https://github.com/tr4wzified/wabbajack/commits/starfield-ba2/
- Download: https://mega.nz/file/4GB3wCTS#6vGbQuKz4rIOE4PRPXCHmna27-_SvbXiUzTbvlGONqc

PR #2589, opened 2024-06-17
Add Starfield support

tr4wzified:
Fixes the issues when trying to compile a Starfield modlist, as the ba2 files have a slightly different format.

Changes include
- BA2 header now includes two unknown uints & compression flag.
- When this compression flag is 3, LZ4 block compression is used on the file contents

Closes https://github.com/wabbajack-tools/wabbajack/issues/2587


EzioTheDeadPoet:
something in the compiler test failed on the windows test 😓

PR #2596, opened 2024-07-08
Use the default file manager when opening a folder.

EzioTheDeadPoet:
#2592


EzioTheDeadPoet:
still need to add the changelog

tr4wzified:
tested

PR #2598, opened 2024-07-13
Use system TMP directories instead of hard coding

majcosta:
Some users might set their TMP (and TEMP) environment variables somewhere other than C: for various reasons

This is set by default in all Windows install so it doesn't break anything for anyone using it


majcosta:
stale

Issue #2571, opened 2024-05-30
Fatal exception on double clicking browser window.


kasanryukin:
**Describe the bug**

Double clicking on the URL bar in the nexus browser window causes a fatal exception with DragMove.

**Additional Context**

Wabbajack specific I think, I don't have nexus premium so I slog through the slow download method.

**Wabbajack Version**

3.6.1.0

**Logs**

00:09:25.278 [ERROR] (Wabbajack.MainWindow) Uncaught error|System.InvalidOperationException: Can only call DragMove when primary mouse button is down.
   at System.Windows.Window.DragMove()
   at Wabbajack.BrowserWindow.UIElement_OnMouseDown(Object sender, MouseButtonEventArgs e) in C:\oss\wabbajack\Wabbajack.App.Wpf\Views\BrowserWindow.xaml.cs:line 46
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.RaiseEventImpl(DependencyObject sender, RoutedEventArgs args)
   at System.Windows.UIElement.RaiseTrustedEvent(RoutedEventArgs args)
   at System.Windows.Input.InputManager.ProcessStagingArea()
   at System.Windows.Interop.HwndMouseInputProvider.ReportInput(IntPtr hwnd, InputMode mode, Int32 timestamp, RawMouseActions actions, Int32 x, Int32 y, Int32 wheel)
   at System.Windows.Interop.HwndMouseInputProvider.FilterMessage(IntPtr hwnd, WindowMessage msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Interop.HwndSource.InputFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
[...]

PR #2606, opened 2024-08-13
 Fix for crash when right clicking or doubler clicking browser url or window bar.

tr4wzified:
Fixes https://github.com/wabbajack-tools/wabbajack/issues/2571

The current code ran Dragmove when any mousebutton event was fired, even right click, and Dragmove crashes when left button isnt held down when it is called.

Added just a guard to check left button is pressed before calling dragmove.

The crashing on double click couldve been caused by unlucky timing or threaded crap causing the function to be called when the button had already been released.

Confirmed crashing behavior prior to change, and confirmed no longer present afterwards

Credit goes to @JanuarySnow


tr4wzified:
tested myself, seems fine 👍

PR #2621, opened 2024-09-22
Fix In-App Gallery not reflecting the automated maintenance

### LINKED ISSUE #2620: In-App Gallery out of sync with the maintenance status on the website.
**Describe the bug**

Gallery shows lists that should be in maintenance as available. 
(Website Gallery works.)

**Wabbajack Version**

3.7.3.0

EzioTheDeadPoet:
Fixes #2620 .

### DIFF
--- a/Wabbajack.App.Wpf/View Models/Gallery/ModListGalleryVM.cs
+++ b/Wabbajack.App.Wpf/View Models/Gallery/ModListGalleryVM.cs
@@ -232,11 +232,12 @@ namespace Wabbajack
             try
             {
                 var modLists = await _wjClient.LoadLists();
+                var modlistSummaries = await _wjClient.GetListStatuses();
                 _modLists.Edit(e =>
                 {
                     e.Clear();
                     e.AddOrUpdate(modLists.Select(m =>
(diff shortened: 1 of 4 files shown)


Wabbajack.App.Wpf/View Models/Gallery/ModListGalleryVM.cs
```diff
@@ -232,11 +232,12 @@ private async Task LoadModLists()
             try
             {
                 var modLists = await _wjClient.LoadLists();
+                var modlistSummaries = await _wjClient.GetListStatuses();
                 _modLists.Edit(e =>
                 {
                     e.Clear();
                     e.AddOrUpdate(modLists.Select(m =>
-                        new ModListMetadataVM(_logger, this, m, _maintainer, _wjClient, _cancellationToken)));
+                        new ModListMetadataVM(_logger, this, m, _maintainer, modlistSummaries, _wjClient, _cancellationToken)));
```
EzioTheDeadPoet, line 240: I could have also added the network call inside each ModListMetaDataVM but it seemed better to reduce the network traffic and instead add it as an additional parameter to the method.

PR #2624, opened 2024-09-29
Update StandardInstaller.cs - Typo in MissingManuals.html generation

code-syl:
Small typo fixed in MissingManuals.html generation.


EzioTheDeadPoet:
Hi just a headsup this workflow will fail since it won't have access to the nexus api key. 
If you taget https://github.com/wabbajack-tools/wabbajack/tree/%232620-fix-in-app-gallerymaintenance instead of main then I can merge it into my open pull request where it will run with the proper api key accessible.

Issue #2625, opened 2024-10-09
Ensure changing the NXM links to match game IDs doesn't break Wabbajack's client


Pickysaurus:
**Describe the bug**
Currently, several NXM links use a different ID to the game domain found on the website. 

| Game | Game ID | NXM ID |
|--------|--------|--------|
| Skyrim SE| `skyrimspecialedition` | `SkyrimSE` |
| Fallout NV | `newvegas` | `falloutnv` | 
| TESO | `elderscrollsonline` | `teso` |

We want to align these with the Game IDs but ensure the popular apps that use NXM links won't be impacted. 

**Additional Context**
https://github.com/Nexus-Mods/Vortex/issues/16552

tr4wzified:
Hey, thanks for notifying us about this. I'll have a look later today to see if this affects WJ :)

PR #2626, opened 2024-10-09
Support NXM links using Nexus game IDs (#2625)

JonathanFeenstra:
When handling NXM links, games are now also looked up by Nexus game ID as a fallback instead of only by MO2 archive name to address #2625. When this change is actually implemented by Nexus, the lookup by MO2 archive name should probably be removed altogether.


Pickysaurus:
Please let me know when this goes into a main release. We would like to track usage to ensure to majority of users are updated before we make the change.

EzioTheDeadPoet:
I will merge this into a dummy branch and test a local build. And then make a new pull request from that.

PR #2651, opened 2024-11-22
exclude mediafire from list validation

iAmMe27:
Exclude mediafire from list validation as it isn't necessarily reliable so it often brings modlists down when not needed.

Note, with this change that modlist authors will need to keep an eye on Mediafire hosted mods to ensure their availability.


EzioTheDeadPoet:
don't be alarmed the test will fail on pull requests that don't come from the main repository.
since there is no way to read the api key in those.

iAmMe27:
All good, I don't fully understand GitHub actions yet anyway, I need to do me some more reading on it

Issue #2664, opened 2024-12-17
WabbaJack CLI set-nexus-api-key not usable


JerryYOJ:
**Describe the bug**

running >wabbajack-cli.bat set-nexus-api-key -k QJ.....== on a fresh install gives error

Unhandled exception: System.InvalidOperationException: A suitable constructor for type 'Wabbajack.Services.OSIntegrated.TokenProviders.NexusApiTokenProvider' could not be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.

Win10

**Wabbajack Version**

3.7.5.1

**Logs**

Unhandled exception: System.InvalidOperationException: A suitable constructor for type 'Wabbajack.Services.OSIntegrated.TokenProviders.NexusApiTokenProvider' could not be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.CreateConstructorCallSite(ResultCache, ServiceIdentifier, Type , CallSiteChain)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.TryCreateExact(ServiceDescriptor, ServiceIdentifier, CallSiteChain, Int32)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.TryCreateExact(ServiceIdentifier, CallSiteChain)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.CreateCallSite(ServiceIdentifier serviceIdentifier, CallSiteChain callSiteChain)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.GetCallSite(ServiceIdentifier, CallSiteChain)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.CreateServiceAccessor(ServiceIdentifier serviceIdentifier)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey, Func`2)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(ServiceIdentifier, ServiceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.GetService(Type)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService[T](IServiceProvider )
[...]

PR #2673, opened 2024-12-28
Remove deprecated nexus auth method

AlexDickerson:
Removing this auth method per: https://github.com/wabbajack-tools/wabbajack/issues/2664


tr4wzified:
This should've been deleted earlier, thanks

PR #2676, opened 2024-12-29
Faster resumable downloads

AlexDickerson:
Replaces the existing ResumableDownloader with a new implementation that uses native HttpClient rather than the Downloader library.

Reworked how downloads are resumed. If a download is found but does not hash as expected, indicating it was not completed, the length of the file is taken as the starting position for a new download.

Progress reporting working as normal. Adjusted to only report progress on files over a certain size in order to eliminate needless work for the UI.

Added a threshold to the log stream shown in the installation UI. Now only logs of Error or above will be displayed. This significantly improves the responsiveness of the UI.

Removed performance settings that are no longer in use.

Impact
- Much faster sustained download speeds, easily capable of saturating a 1gb/s connection.
- Much lower memory usage during downloads
- Much more responsive UI. UI freezes should be totally eliminated

This replaces this PR: https://github.com/wabbajack-tools/wabbajack/pull/2676#issue-2762163954


Wabbajack.App.Wpf/Models/LogStream.cs
```diff
@@ -59,6 +59,8 @@ protected override void Dispose(bool disposing)
 
     protected override void Write(LogEventInfo logEvent)
     {
+        if(logEvent.Level < LogLevel.Error)
+            return;
```
tr4wzified, line 63: The log is referred to a lot when checking what's going on, not sure about this. Maybe better suited for a separate PR to discuss this.

Wabbajack.App.Wpf/Wabbajack.App.Wpf.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <OutputType>WinExe</OutputType>
+    <OutputType>Exe</OutputType>
```
tr4wzified, line 4: Any reason for this?

Wabbajack.Installer/AInstaller.cs
```diff
@@ -370,29 +370,28 @@ public async Task DownloadMissingArchives(List<Archive> missing, CancellationTok
...
-                        var uniqueKey = archive.State.PrimaryKeyString.StringSha256Hex();
-                        outputPath = _configuration.Downloads.Combine(origName + "_" + uniqueKey + "_" + ext);
-                        outputPath.Delete();
-                    }
+        var missingBatches = missing.Batch(100).ToList();
 
-                var hash = await DownloadArchive(archive, download, token, outputPath);
-                UpdateProgress(1);
-            });
+        List<Task> downloadBatches = [];
```
tr4wzified, line 376: I think the PDoAllBatched method is more suited for batching here

Wabbajack.Networking.Http/ResumableDownloader.cs
```diff
@@ -1,187 +1,146 @@
...
+        }
+        catch (Exception ex) when (ex is SocketException || ex is IOException)
+        {
+            _logger.LogWarning(ex, "Failed to download '{name}' due to network error. Retrying...", filePath.FileName.ToString());
+
+            if (retry == 0)
             {
-                _logger.LogWarning("Download for '{name}' doesn't support resuming. Deleting package...", _outputPath.FileName.ToString());
-                DeletePackage();
+                throw;
```
tr4wzified, line 57: Would be nice to keep a log message here so we're sure it threw in this spot

Wabbajack.Networking.Http/ResumableDownloader.cs
```diff
@@ -1,187 +1,146 @@
...
         }
-    }
 
-    private void OnDownloadStarted(object? sender, DownloadStartedEventArgs e)
-    {
-        _job.ResetProgress();
+        var responseStream = await response.Content.ReadAsStreamAsync(token);
 
-        if (_job.Size < e.TotalBytesToReceive)
+        long reportProgressThreshold = 100 * 1024 * 1024; // Report progress every 100MB
```
tr4wzified, line 105: Smart, those small files don't really matter anyway. However 100MB might be kind of large still, for those with slower connections. Maybe better to lower this to say 10MB?

Wabbajack.Networking.Http/ResumableDownloader.cs
```diff
@@ -1,187 +1,146 @@
...
+        bool shoudlReportProgress = job.Size > reportProgressThreshold; // Reporting progress on small files just generates strain on the UI
```
tr4wzified, line 106: Small typo here

tr4wzified (commented):
Awesome stuff, just a few remarks. Been a while since we had another contributor showing up :)

Wabbajack.App.Wpf/Models/LogStream.cs
AlexDickerson: Yeah, this aught to be out of scope. Will remove and suggest separately.

Wabbajack.App.Wpf/Wabbajack.App.Wpf.csproj
AlexDickerson: I change that to get the console to popup. Will revert once PR is published.

If I can get the console without mucking with the proj file lemme know : )

Wabbajack.Installer/AInstaller.cs
AlexDickerson: That shouldn't be needed anymore. The 10 min cooldown Nexus kept hitting me with seems to have been from the repeated requests to the validate endpoint, which has been fixed. 

Will revert this change.

Wabbajack.Networking.Http/ResumableDownloader.cs
AlexDickerson: Added

Wabbajack.Networking.Http/ResumableDownloader.cs
AlexDickerson: Adjusted.

I think there's some good improvements that could be made to managing settings, could easily make that an adjustable value if we valued it enough.

Wabbajack.Networking.Http/ResumableDownloader.cs
AlexDickerson: Fixed

Wabbajack.App.Wpf/Wabbajack.App.Wpf.csproj
tr4wzified: I assume you use Visual Studio? You can see this in the Output window at the bottom

tr4wzified (approved):

Wabbajack.App.Wpf/Wabbajack.App.Wpf.csproj
AlexDickerson: Figured I should be able to, just haven't been seeing it. I've probably hidden a window or something at some point.

Wabbajack.Installer/AInstaller.cs
tr4wzified: It's under the Debug menu at the top > Windows > Output

PR #2677, opened 2024-12-29
Filter Logs in ModList intaller view

AlexDickerson:
Currently all logs get displayed in the ModList installer. This causes some significant decreases in responsiveness. It also means that more meaningful messages like download failure are lost behind logs that aren't meaningful to most users.

This filters to warnings and above to reduce the noise and improve UI performance.

### DIFF
--- a/Wabbajack.App.Wpf/Models/LogStream.cs
+++ b/Wabbajack.App.Wpf/Models/LogStream.cs
@@ -68,6 +68,7 @@ public class LogStream : TargetWithLayout
         string ShortMessage { get; }
         DateTime TimeStamp { get; }
         string LongMessage { get; }
+        LogLevel Level { get; }
     }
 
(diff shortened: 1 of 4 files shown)


tr4wzified:
Personally not the biggest fan of this. I know for the average user this log might not say much, but in the new UI the log isn't visible by default anyway. Also, this log view is used on compilation where I think the author would definitely prefer a more verbose log about what's going on while things are happening. Is it that much faster if you log less info to the control? If there is a serious performance issue there I would rather fix that instead of just showing less info.

AlexDickerson:
The UI is significantly more responsive with fewer log statements being written out. There very likely is a way to improve performance but my knowledge of xaml/wpf stuff is not sufficient to know where to begin. 

Maybe the filter could be exposed as a option somewhere in the UI?

AlexDickerson:
Looking forward to checking out the new UI : )

PR #2681, opened 2025-01-08
Prototype to use ManualBlobDownload for LL

zclimber:


EzioTheDeadPoet:
this should probably looked at again 🤔 
(if it doesn't break with any of the many changes made in 4.0)

PR #2686, opened 2025-01-19
Consolidated Downloaders into one project

AlexDickerson:
This PR seeks to reduce the quantity of projects in the current solution. It consoldiates the various Wabbajack.Downloaders.* into a single project using folders for organization.

It moves the affected files into a single root Wabbajack.Downloaders folder.

The only changes to the content of the affected files are
1. Edits to .csproj files to accomodate the new single project
2. ServiceExtensions combined into a single class in Wabbajack.Downloads
3. Using statements adjustd accordingly
4. NexusDownloaer.cs adjusted to specify namespace of "Manual" due to name collision.


tr4wzified:
Sorry I didn't respond to this sooner, but not a big fan of this. I prefer sticking to a more modular approach. Package management can now be managed per project, that way no downloaders will have unnecessary dependencies.

PR #2708, opened 2025-03-18
Add additonal checks to GetGPUName() to avoid issues with virtual GPUs

kaeltis:
As discussed on discord (https://discord.com/channels/605449136870916175/640958797820461082/1350043639228928000), this should fix the following issues in `Wabbajack.Util.GetGPUName()`:
 - Exceptions with (virtual) GPUs that have no refresh rate or description by validating the objects before casting them
 - Returning virtual GPUs by filtering on common manufacturers
 - Always returning the last GPU instead of the one with the highest refresh rate

### DIFF
--- a/Wabbajack.App.Wpf/Util/SystemParametersConstructor.cs
+++ b/Wabbajack.App.Wpf/Util/SystemParametersConstructor.cs
@@ -124,7 +124,7 @@ namespace Wabbajack.Util
 
         private string GetGPUName()
         {
-            var gpuManufacturers = new [] {"amd", "intel", "nvidia"};
+            HashSet<string> gpuManufacturers = ["amd", "intel", "nvidia"];
             string gpuName = "";
             uint gpuRefreshRate = 0;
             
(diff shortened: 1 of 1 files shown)


Wabbajack.App.Wpf/Util/SystemParametersConstructor.cs
```diff
@@ -124,18 +124,24 @@ public SystemParameters Create()
...
-                    var currentRefreshRate = (uint)obj["CurrentRefreshRate"];
-                    if (currentRefreshRate > gpuRefreshRate)
-                        gpuName = obj["Description"].ToString();
+                    if(obj["CurrentRefreshRate"] != null && obj["Description"] != null) {
+                        var currentRefreshRate = (uint)obj["CurrentRefreshRate"];
+                        var currentName = obj["Description"].ToString();
+                        
+                        if (gpuManufacturers.Any(s => currentName.Contains(s, StringComparison.OrdinalIgnoreCase)) && currentRefreshRate > gpuRefreshRate)
+                            gpuName = currentName;
+                            gpuRefreshRate = currentRefreshRate;
```
tr4wzified, line 143: If you leave out the brackets after an if, only the next statement will be executed. Meaning the GpuRefreshRate will be updated on every cycle

Wabbajack.App.Wpf/Util/SystemParametersConstructor.cs
```diff
@@ -124,18 +124,24 @@ public SystemParameters Create()
 
         private string GetGPUName()
         {
+            var gpuManufacturers = new [] {"amd", "intel", "nvidia"};
```
tr4wzified, line 127: Any particular reason to filter on these? Would also use a HashSet here for faster lookups. Even if it's a super small array, it's good practice when you want to have a list that you're gonna do lots of Contains on

tr4wzified (changes requested):
Added some comments. I haven't been able to look into whether there's a better way to determine the primary GPU yet, are you sure it's always the last one returned by Windows?

tr4wzified (approved):

tr4wzified:
Sorry for not paying attention earlier, busy with too many different things at the same time x)
The HashSet advice I gave doesn't really apply in this case since we're doing a contains on a string, not on that list we're checking. Anyway, doesn't matter in this case

PR #2711, opened 2025-03-18
Disable Windows tests since they're unreliable

tr4wzified:


tr4wzified:
nvm, it fixed itself already

PR #2740, opened 2025-03-22
Publish Button visual progress

JanuarySnow:
Its supposed to update the text and percentage as it goes along, but it dosnt, idk why.

But at least it disables the button and shows something happening and confirms when its done.

please tell me why the UI element isnt updating 

https://github.com/user-attachments/assets/69f4f23a-1f37-4d11-851e-99d69ca9dcea


Wabbajack.App.Wpf/ViewModels/Compiler/CompilerMainVM.cs
```diff
@@ -138,14 +147,66 @@ private async Task Publish()
...
+            try
+            {
+                var (progress, publishTask) = await _wjClient.PublishModlist(
+                    Settings.MachineUrl, System.Version.Parse(Settings.Version), Settings.OutputFile, downloadMetadata);
+
+                var progressSubscription = progress
+                    .ObserveOn(RxApp.MainThreadScheduler)
+                    .Subscribe(p =>
+                    {
+                        RxApp.MainThreadScheduler.Schedule(_logger, (_, _) =>
```
tr4wzified, line 179: You're already on the UI thread here by doing ObserveOn(RxApp.MainThreadScheduler). I would replace this with ObserveOnGuiThread to be consistent across the application (it's the same, it's just a WJ-specific shorthand for it)

Wabbajack.Networking.WabbajackClientApi/Client.cs
```diff
@@ -548,47 +550,67 @@ public async Task<IReadOnlyList<string>> GetMyModlists(CancellationToken token)
...
-        _logger.LogInformation("Uploading modlist {MachineUrl}", namespacedName);
-        
+
+        var uploadProgress = new double();
         var (progress, uploadTask) = await UploadAuthorFile(modList);
-        progress.Subscribe(x => _logger.LogInformation(x.Message));
-        var downloadUrl = await uploadTask;
         
-        _logger.LogInformation("Publishing modlist {MachineUrl}", namespacedName);
+        var disposable = progress.Subscribe(m =>
```
tr4wzified, line 569: Instead of doing this, I would take a look at the Subject used in UploadAuthorFile, that way you don't have to manually mess around with the observable and a subscribe

tr4wzified (changes requested):
Left some comments. Do you think you can get further with the publish button progress yourself, or would you prefer I get into it myself?

JanuarySnow:
cleaned up and the UI element is updating now

PR #2820, opened 2025-06-27
Fix Wabbajack CLI for hash-game-files verb

JanuarySnow:
Seems to work and provide exact same results as the old one did. 

I may have used some helpers and utils incorrectly as I re-implemented this as I didnt understand a lot of the associated code, but writing to JSON aint exactly rocket science in the end.


tr4wzified:
Looks good to me 👍Thanks

PR #2823, opened 2025-06-29
[WIP] Add ability to extract EXE files

### LINKED ISSUE #2822: .EXE extraction support

Some mods use installers making their assets unsourceable.

WJ can handle some/most exe installers natively.

Authors add manunal instructions to install installer mods. 

**Additional context**

Goal is to implement https://constexpr.org/innoextract/ the same way 7zip is implemented. 

Using `-i` to identify `.exe` files that can't be extracted and should be skipped.

EzioTheDeadPoet:
closes #2822.

This pull request also replaces the GUID's that are used for Temporary File Paths with [shortid](https://github.com/bolorundurowb/shortid)'s to prevent the `MAX_PATH` issues on windows.


EzioTheDeadPoet:
This would still need the implementation with the GUI, and ideally some unit test. 

And there is still the open question of moving the temp folder closer to a drive because it can still occasionally happen that we hit the file path limit. (My debug builds still do while compiling because the temp folder is in the folder where WJ is located, while the temp folder for installs is currently in the installation directory.)

And so far the only failed extractions I had while debugging and testing this were due to paths that were too long.

Here a log of my successful compile of an abomination test list (JTKs Cyberpunk MO2077 list with an extracted inno installer in the files and sources in addition to the og list): [Wabbajack.current.log](https://github.com/user-attachments/files/20970515/Wabbajack.current.log)

The `.wabbajack`: [Modlist_Output.zip](https://github.com/user-attachments/files/20970518/Modlist_Output.zip)

And a log of a list update/installation (I kept everything and only deleted the extracted installer files to save time because the extraction is what needed to be tested): [Wabbajack.current.log](https://github.com/user-attachments/files/20970539/Wabbajack.current.log)


Yes the logs contain a lot of extraction information rn but I needed that to keep my sanity while debugging 😓 .
Ideally we would probably have the indivial files be silent and just parse the log output in another way to get some statistics to display on the GUI.

@tr4wzified @JanuarySnow I am not sure how the GUI stuff is setup and how I could use it. 
I could look into ways how we could work on configuring and reading the innoextractor log to query some good information for the GUI to use (maybe we need to run the `--list`  option first to get the file sizes printed in a log so we can map them to the actual extraction log and work with that, if the actual extraction logging can't be configured to show file sizes that is) but rn I wouldn't know where and how to forward any of that to the GUI. 

 (@daulvec "@ ed" so you are up to date on the current status, as it stands rn it looks like I could send you a very WIP build that has some very obvious flaws in the GUI and logging departments in the next days so you could get testing on an actual list for a game where this will be useful.)

EzioTheDeadPoet:
I tried different log output configurations `innoextract.exe` is capable (so we have more than just the line count to go off).
And I found something very interesting.

Since it has no prediction for the size of all files being extracted we need to calcualte that ourselfs, but ideally before we start the actual extraction. The tool has a `--list-sizes` mode where it just prints the files and their individual extracted sizes. 

That command outputs the following log, which will be a bit more complicated for us to parse for adding up a estimated total extraction size to be used for a potential progess bar:  [formatted.log](https://github.com/user-attachments/files/20983702/formatted.log)

Now the tool has another flag `--silent` that that has a way better output where it only outputs the size in bytes and the file name, which looks like this: [formatted-silent.log](https://github.com/user-attachments/files/20983726/formatted-silent.log)

Now the problem is that the better formatted for parsing (the one run with `--silent`) option takes a minute or longer to write (depends on the installer), while the less ideal output is done in a few seconds instead.
So it seems better suited for generating comparison data fast so we can compare the extraction state with the max amount of data to be extracted in the actual extraction call. But this means we would need to implement functions to process each line of that log ourselfs and hope that we are faster than it would be to let `innoextract.exe` do the job. 

And then we also need to process the actual extraction log (which we can force to look identical so we should be able to use same parsing options).

As for an overal percentage counter we could either make it depend on the lines aka number of files (since we will need a dry run anyway to collect the data for the UI) or the file sizes.

Unless we just make it an infinite "spinner" bar, that doesn't fill but has little blobs moving from left to right until the file is exracted, which would save us more complex implementations to gather more detailed data for the UI and would likely end up with better performance since could also ignore all the logging inside `innoextract` and reduce the workload on that process.

tr4wzified:
The exe extraction looks alright from a quick glance, but I'm not convinced about shortid yet. It's quite a major change as it's going to affect extraction of all temporary files, not just EXEs - so I really want to be sure about this before changing something like this.

It looks to me like it's kind of a random library with not a lot of usage, no docs on when we should start seeing possible collisions or anything, no performance benchmarks on how it compares. Can we guarantee we won't get any collision issues with just 8 characters?

EzioTheDeadPoet:
> The exe extraction looks alright from a quick glance, but I'm not convinced about shortid yet. It's quite a major change as it's going to affect extraction of all temporary files, not just EXEs - so I really want to be sure about this before changing something like this.
> 
> It looks to me like it's kind of a random library with not a lot of usage, no docs on when we should start seeing possible collisions or anything, no performance benchmarks on how it compares. Can we guarantee we won't get any collision issues with just 8 characters?

The library is not popular and if you want I can do a little dive how it works under the hood to ensure no duplicates. 
But with the configuration I set rn a-Z (`26*2`(cause lowercase and uppercase characters)) and 0-9 (`10`) and 8 characters per ShortId we get `(26*2+10)^8=218.340.105.584.896` or `2,18340105584896e+14` possible character combinations which for WJ should be enough especially since this is in 2 nested folders generated like that. (Yes I estimated it before but now thought I probably should have done the proper math before that.)
And iirc duing installs (or compiles) one of the "parent" temp folders actually adds the process ID infront of the string for the folder as well. (Forgot which step it was over the week of not looking at this but it can be referenced in the log.)

But based on the math if the lib does enough checks to prevent duplicate assignements it should be fine. 
I would probably have coded something similar but less deadicated to get rid of the length of GUIDs. 
So I thought using a lib should save some time there.

The worst case I could see is a little perfomance decrease from using it over using GUIDs which are a tested reliable standard, because it probably needs to cache and check the IDs it generates against the ones in use (will check that soon). 
But I would consider that potential tiny performance decrease negligible, especially in the context of reducing the chances for file path crashes (compile and installation failures).

EzioTheDeadPoet:
Ok I let an online calculator for the "Birthday Problem" do the math for me cause it is matches the use case.
And to even have a ``0,00001%` chance of getting a duplicate we would need to generate `270267` temp folders with unique identifiers. 
It is not 0 but considering that we at most have CPU-Threats amount of folders active at a time this should not be an issue. 

Now we could reduce that chance to 0 by saving and comparing each `ShortId` before it gets used but looking at those chances that cold potentially be overkill. (Tho it would reduce the chance of colision to aboslute 0 which technically it wasn't with GUIDs either :D )

image
Source Website Calculator: https://www.bdayprob.com/
Source Python implementation/backend: https://github.com/fast-reflexes/BirthdayProblem-Python

EzioTheDeadPoet:
> I realize now WJ actually doesn't have quite as many temporary folders and files in place at the same time as I thought, did some installs and it maxxed out at like 250. That's not equal to the amount of threads used, but I'm no longer worried about collisions. The numbers there are much lower than I thought. I wouldn't do an equals check, it's fine. I'm gonna do a quick performance test.

Ah yeah I think the cleanup is a bit slower in some cases (especially when installing where it might needs the files more than once from the same archive) and we have multiple stages that can start in parallel and make temp folders.

But yeah definitely not enough to cause issues.

Only thing the feature is still missing is the proper UI integration now. So that users don't assume WJ is frozen.

tr4wzified:
Gotta head out for a bit, but I finished the benchmark. It's slower, but I don't think it will be an issue.

https://gist.github.com/tr4wzified/b5470c9464f2867ca13129377da0ce1f

EzioTheDeadPoet:
> Oh shit lol, I just wrote some code to fix the UI not being updated but I was reading back a bit and didn't notice that silent log. That's a bit faster and easier to parse. I'll take a look at using that instead tomorrow.

Don't, the silent log takes longer to print and slows down `innoextract.exe` significantly. 
(Or at least don't use it unless you did some benchmarks first cause I would estimate that your code will be faster than using the silent log.)

tr4wzified:
Ah that's right. Late night coding gets my mind even more scattered, and it's bad enough as it is x) I'll get everything merged in the 4.0.3.0 branch tomorrow then, does this look fine to you as well? I benchmarked my code, the helper takes about 820ns to parse your extraction log, it's pretty fast.

EzioTheDeadPoet:
We probably can delete my debug logging that logs out each individually extracted file in the written log cause that will be a bit excessive in a release and makes logs harder to read but besides that I tested this branch on my end with the test install and it worked flawlessly 🎉 
Thanks for taking over the work of paring the logging data into the UI elements. 😄

tr4wzified:
Oh yeah, we should definitely take that debug logging out. Thanks to you too for doing all the research, it was already 90% there :)

PR #2824, opened 2025-06-30
Fixed the automated version number with signed builds

EzioTheDeadPoet:
Changed the configs to use the `VERSION` environment variable again that is found via the `CHANGELOG.md` file by the `build_all.bat` build script that is used for making the signed release builds.    

I also added a fallback for debug builds to use `4.0.0.0` as their version number if the environment variable is not set on the current system.


EzioTheDeadPoet:
I also updated the Copyright field for the main app (the only project file that defines it) to automatically grab the current year of the build machine, cause we were still stuck in 2024 with that.

tr4wzified:
Nice work

PR #2842, opened 2025-09-06
feat: Integrate uBO Lite ad-blocking functionality

sheepdestroyer:
Integrates uBO Lite's ad-blocking capabilities directly into Wabbajack's internal web browser.

This is achieved by:
- Creating a new `AdBlockService` to manage and apply ad-blocking rules.
- Using the `DistillNET` library to parse AdBlock-style filter lists.
- Downloading filter lists from uBlock Origin's uAssets repository and EasyList at application startup.
- Intercepting network requests in all `WebView2` instances and blocking them if they match the filter rules.

This enhances user experience by blocking ads and trackers in the various web views used for logins and manual downloads.

### DIFF
--- a/Wabbajack.App.Wpf/App.xaml.cs
+++ b/Wabbajack.App.Wpf/App.xaml.cs
@@ -71,7 +71,7 @@ public partial class App
         var webview2 = _host.Services.GetRequiredService<WebView2>();
 
         var adBlockService = _host.Services.GetRequiredService<Services.AdBlockService>();
-        adBlockService.Initialize();
+        _ = adBlockService.Initialize();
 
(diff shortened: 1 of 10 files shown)


JanuarySnow:
apparently Nexus will limit download speed by half if it detects an adblocker , so for free users, who are the most likely to need this, it will cripple their download speed

For other sites using a manualurl then I can see the utility, but if there is perhaps a way of not using it for nexus links thatd be useful

sheepdestroyer:
Oh. 
Ok, didn't know.
I'll drop that then

sheepdestroyer:
dropped

PR #2931, opened 2026-06-30
Bump Nexus upload-action to v1.0.0-beta.8 for v3 API

Jack97:
Renames file_group_id->file_id and file_category->category per the beta.8 breaking changes. Required before the old upload endpoint is removed on 2026-09-09.


halgari:
Thanks for the help!

PR #2952, opened 2026-08-18
fix for saves that aren't just plain files but more complex nested folders

EzioTheDeadPoet:
To anyone I assigned as a reviewer this is mainly to get at least a second set of eyes on this


EzioTheDeadPoet:
closes #2952

PR #3030, opened 2026-09-20
Log in to Nexus Mods in the user's own browser

halgari:
The Nexus login ran in an embedded WebView2 and caught the redirect by watching that window try to navigate to `https://127.0.0.1:1234`, which nothing was ever listening on. It had a cookie jar of its own, so a browser already signed in to Nexus did not help, and it put a password field inside the app.

It now opens the system browser and listens for the redirect.

`NexusOAuthLogin` (in `Wabbajack.Networking.NexusApi`) runs the flow: PKCE, the authorize URL, the state check, the token exchange, and what gets stored. `LoopbackOAuthCallback` opens one `TcpListener` on `IPAddress.Loopback` at port 0, so the OS picks the port and nothing off the machine can reach it. The redirect URI is read off the listener — `http://127.0.0.1:<port>/oauth/callback` — rather than being a constant, since the authorize request and the token exchange both have to repeat whatever port it turned out to be.

Details worth knowing:

- Only `/oauth/callback` is answered; everything else on the port is a 404, because a browser will ask for a favicon.
- Each connection is handled on its own, because browsers open speculative connections and send nothing down them. One of those must not hold the port while the real redirect waits behind it.
- A connection is half-closed and drained before it is dropped. Only the request line is read, and closing a socket with the rest of the request still in its buffer is an RST, which the browser shows in place of the page just written to it.
- `HttpListener` would have been shorter and is deliberately not used: it hands the prefix to http.sys, which is a machine-wide registration with its own ACLs, so a failure there is an elevation prompt rather than a bad login.
- `state` is checked before the code is spent, since anything on the machine can reach a loopback port.
- `IOAuthBrowser` is the seam that leaves the process, `TryAdd`ed so a host or a test can register its own.

`StateToStore` comes across unchanged — a refusal from the token endpoint still writes nothing — and its test table moves to `Wabbajack.Networking.NexusApi.Test` with the code. `LoopbackOAuthCallbackTests` covers the redirect, a refusal, ignored paths, a silent connection, cancellation closing the port, and that only loopback is bound. `NexusLoginHandler` and its base64 helper are deleted; the settings tile and the `nexus-login` verb call the flow directly, both with a ten-minute timeout, which is what closes the port when nobody finishes a login.

Offline lane passes, build is warning-free.

**Needs checking against Nexus:** the authorize request now sends `http://127.0.0.1:<ephemeral>/oauth/callback` where it used to send `https://127.0.0.1:1234`. That is the RFC 8252 loopback rule, but it only works if the `wabbajack` OAuth client is registered to accept loopback redirects on a variable port over plain http. If Nexus exact-matches the old value, the authorize step will come back `invalid_request` and the client registration has to be updated on their side.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VMhysLPMUEDGv76gxGtZE6


halgari:
Superseded by #3027, which removes WebView2 across the app and includes the same loopback Nexus login in a more decomposed form (OAuthLoopbackListener, OAuthQuery, CallbackPage). #3027 also carries three fixes this branch does not: it starts listening before the browser is opened, bounds the token exchange, and parses request lines the same way on Unix as on Windows. The one behaviour I thought was unique here - half-closing the callback socket before disposing it, so the browser shows the success page rather than a connection reset - is already in #3027. Nothing left to rebase.

